PR #35, opened 2020-09-11
Kimmy/dpe rr

jya-kmu:
Duplication is not handled by this commit. Just push the temporary round-robin method into dpe for now, which is still upon change.

### DIFF
--- a/src/data_placement_engine.cc
+++ b/src/data_placement_engine.cc
@@ -100,12 +100,12 @@ Status RoundRobinPlacement(SharedMemoryContext *context, RpcContext *rpc,
 
       // Construct the vector for the splitted blob
       std::vector<size_t> new_blob_size;
-      size_t bolb_each_portion {blob_sizes[i]/split_num};
+      size_t blob_each_portion {blob_sizes[i]/split_num};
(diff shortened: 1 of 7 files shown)


src/data_placement_engine.cc
```diff
@@ -58,6 +58,102 @@ Status TopDownPlacement(SharedMemoryContext *context, RpcContext *rpc,
...
+      std::vector<int> split_choice(split_range, split_range+split_option-1);
+
+      // Random pickup a number from split_choice to split the blob
+      std::uniform_int_distribution<std::mt19937::result_type>
+        position(0, split_choice.size()-1);
+      int split_num = split_choice[position(rng)];
+
+      // Construct the vector for the splitted blob
+      std::vector<size_t> new_blob_size;
+      size_t bolb_each_portion {blob_sizes[i]/split_num};
```
ChristopherHogan, line 103: Typo "bolb"

ChristopherHogan (approved):

src/data_placement_engine.cc
jya-kmu: Thanks!

PR #43, opened 2020-10-29
quick fix for random placement bug

jya-kmu:
quick fix


ChristopherHogan:
The `RandomPlacement` algorithm isn't quite right. Consider the failing test case. We have 4 tiers. The available buffering space for tier 0 is 4096 KB, for tier 1 8 bytes, for tier 2 8 bytes, and for tier 3 8 bytes (I set it up this way to force blobs to go to swap space). Since first blob we're trying to place is 4KB, it is not split. However, the `RandomPlacement` function returns  a `PlacementSchema` that wants to place 4096 KB in tier 1, which is invalid because tier 1 only has 8 bytes of space available.

jya-kmu:
Since node local capacity is used for now, which function should be called to update node local capacity, something like UpdateRemainingNodeCapacities?

ChristopherHogan:
You're right, I'm not updating the node level targets anywhere. I'll create a quick PR.

PR #44, opened 2020-10-30
DPE related bug fixes and refactoring

ChristopherHogan:
Replaces #43.
* Added a single function (`UpdateBufferingCapacities`) that handles all capacity updates. We currently update a global system view state and node level capacities, but the two will likely merge into one once we have topologies and targets worked out.
* Factored `GetValidSplitChoices` and `AddRandomSchema` out of `RandomPlacement` (in the library and in the tests) to eliminate code duplication.
* Corrected behavior of `AddRandomSchema`.
* `RoundRobinPlacement` operates on node state instead of global state.
* Properly initialize `ordered_cap` in `RandomPlacement`.
* `RandomPlacement` updates its local copy of `node_state` (working on outdated information or course).

### DIFF
--- a/src/buffer_pool.cc
+++ b/src/buffer_pool.cc
@@ -426,6 +426,24 @@ static void IncrementAvailableBuffers(SharedMemoryContext *context,
   }
 }
 
+void UpdateBufferingCapacities(SharedMemoryContext *context, i64 adjustment,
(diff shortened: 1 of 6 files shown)


jya-kmu (approved):
I will use this version to avoid conflict.

PR #53, opened 2020-12-11
Set upper limit to name lengths

ChristopherHogan:
Restricting bucket and blob names to 128 bytes so we can get an upper bound for our metadata size.


jya-kmu (approved):
Yes, this limit setup is necessary.

PR #55, opened 2020-12-15
Kimmy/code coverage

jya-kmu:
Enable ctest with code coverage with github action.


CMakeLists.txt
```diff
@@ -141,7 +141,8 @@ endif()
 # Build options
 #------------------------------------------------------------------------------
 option(BUILD_SHARED_LIBS "Build with shared libraries." ON)
-option(HERMES_INTERCEPT_IO "Use GOTCHA as IO interception tool" ON)
+option(BUILD_TESTING "Build testing." ON)
```
ChristopherHogan, line 144: According to [the documentation](https://cmake.org/cmake/help/v3.10/module/CTest.html?highlight=build_testing), the `CTest` module "automatically creates a `BUILD_TESTING` option that selects whether to enable testing support (`ON` by default)."

CMakeLists.txt
```diff
@@ -316,11 +337,9 @@ endif()
 #-----------------------------------------------------------------------------
 # Testing
 #-----------------------------------------------------------------------------
-if(CMAKE_PROJECT_NAME STREQUAL HERMES)
+if(NOT HERMES_EXTERNALLY_CONFIGURED AND BUILD_TESTING)
+  enable_testing()
```
ChristopherHogan, line 341: From [the documentation](https://cmake.org/cmake/help/v3.10/command/add_test.html#command:add_test):
> CMake will generate tests only if the enable_testing() command has been invoked. **The CTest module invokes the command automatically when the BUILD_TESTING option is ON.**

CMakeLists.txt
```diff
@@ -316,11 +337,9 @@ endif()
 #-----------------------------------------------------------------------------
 # Testing
 #-----------------------------------------------------------------------------
-if(CMAKE_PROJECT_NAME STREQUAL HERMES)
+if(NOT HERMES_EXTERNALLY_CONFIGURED AND BUILD_TESTING)
```
ChristopherHogan, line 340: Can you explain what the `HERMES_EXTERNALLY_CONFIGURED` option is for? I've never found it useful for anything.

.github/workflows/main.yml
```diff
@@ -1,4 +1,5 @@
 # This is a basic workflow to help you get started with Actions
+# Dashborad location https://cdash.hdfgroup.org/index.php?project=HERMES
```
ChristopherHogan, line 2: Typo, should be "Dashboard"

CMakeLists.txt
```diff
@@ -153,6 +154,10 @@ option(HERMES_DEBUG_HEAP "Store Heap debug information for visualization." OFF)
 option(HERMES_BUILD_BENCHMARKS "Build the Hermes benchmark suite." OFF)
 option(HERMES_ENABLE_TIMING "Turn on timing of selected functions." OFF)
 
+if(NOT HERMES_EXTERNALLY_CONFIGURED)
+  option(HERMES_ENABLE_COVERAGE "Enable coverage." OFF)
```
ChristopherHogan, line 158: Why is this in an `if` block.

test/CMakeLists.txt
```diff
@@ -25,11 +26,15 @@ foreach(program ${API_TESTS})
     "${CMAKE_BINARY_DIR}/bin/${program}")
   set_tests_properties("Test${program}" PROPERTIES ENVIRONMENT
     LSAN_OPTIONS=suppressions=${CMAKE_CURRENT_SOURCE_DIR}/data/asan.supp)
+  if(HERMES_ENABLE_COVERAGE)
+    set_coverage_flags(${program})
```
ChristopherHogan, line 30: I don't think we need the coverage flags on each test. We're only interested in coverage in the library, not which lines get executed in the tests themselves. Right?

test/script/ref_script.cmake
```diff
@@ -0,0 +1,104 @@
+# Usage:
```
ChristopherHogan, line 1: What is this file for? Is it left over from testing/debugging?

ChristopherHogan:
When I examine the most recent coverage report generated [here](https://cdash.hdfgroup.org/viewCoverage.php?buildid=58828&status=6&dir=src) it only lists 4 files. Where are the rest?

CMakeLists.txt
jya-kmu: Yes, I was thinking to rename it to `HERMES_BUILD_TESTING`, then there will be no confusion.

test/script/ref_script.cmake
jya-kmu: That's for ppl who wants to do ctest locally from a script, which can be used as reference for connivence.

test/CMakeLists.txt
jya-kmu: I have seen both cases, including test or not. Either is fine.

jya-kmu:
> When I examine the most recent coverage report generated [here](https://cdash.hdfgroup.org/viewCoverage.php?buildid=58828&status=6&dir=src) it only lists 4 files. Where are the rest?

Check this one (https://cdash.hdfgroup.org/viewCoverage.php?buildid=58916&status=6&dir=src). The reason is test name inconsistency.

CMakeLists.txt
jya-kmu: Coverage is done only when we are building hermes, but not when using it as a dependency of another library.

PR #58, opened 2020-12-18
Move from DeviceID to TargetID in data placement

ChristopherHogan:
* Changed `PlacementSchema` from `std::vector<std::pair<size_t, DeviceID>>` to `std::vector<std::pair<size_t, TargetID>>`.
* All data placement now works with `TargetID`s instead of `DeviceID`s.
* I made some name changes in the DPE code to help improve my understanding. Let me know if I got anything wrong.
* We still only work with local targets, but this PR provides the infrastructure to support neighborhoods and global lists of targets.

### DIFF
--- a/src/hermes_types.h
+++ b/src/hermes_types.h
@@ -46,11 +46,22 @@ constexpr char kPlaceInHierarchy[] = "PlaceInHierarchy";
 #define HERMES_NOT_IMPLEMENTED_YET \
   LOG(FATAL) << __func__ << " not implemented yet\n"
 
+union TargetID {
+  struct {
+    u32 node_id;
+    u16 device_id;
+    u16 index;
+  } bits;
+
+  u64 as_int;
+};
+
 /**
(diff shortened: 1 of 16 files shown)


src/hermes_types.h
```diff
@@ -46,11 +46,22 @@ constexpr char kPlaceInHierarchy[] = "PlaceInHierarchy";
 #define HERMES_NOT_IMPLEMENTED_YET \
   LOG(FATAL) << __func__ << " not implemented yet\n"
 
+union TargetID {
+  struct {
```
jya-kmu, line 50: Can we have an explanation here how to understand this struct as a `TargetID`?

jya-kmu (approved):
Looks good to me. Thanks!

src/hermes_types.h
ChristopherHogan: I added some documentation. Does that answer your question?

jya-kmu: Yes, thanks!

PR #64, opened 2021-01-12
Hariharan/vbucket trait

hariharan-devarajan:
This PR consists of functionality required for POSIX and STDIO adapter. Currently, It doesn't support shared traits/vBuckets as none of the metadata is stored in shared memory. I believe we will do it as we need it.


ChristopherHogan:
Could you please remove the `*.pyc` files and add `__pycache__/` to `.gitignore`?

ChristopherHogan:
There are a lot of arbitrary white space changes. They really clutter up the PR and make it difficult to focus on the important changes.

hariharan-devarajan:
I have tried to reduce the whitespace a little. Hope that helps.

ChristopherHogan:
Thank you.

ChristopherHogan:
This PR triples the compilation time from ~7 seconds to ~20 seconds. Is there anything you can do to lower that?

hariharan-devarajan:
The compilation unit for Hermes is the same as you recommended. It must be due to the test case which uses catch2 framework.

ChristopherHogan:
> The compilation unit for Hermes is the same as you recommended.

Actually, I recommended that you `#include` `traits.cc` in `buffer_pool.cc` to keep the number of compilation units low, but instead you have `traits.cc` in `HERMES_SRCS` in `src/CMakeLists.txt`, which creates an additional compilation unit. Making that change appears to reduce the time by a few seconds.

If Catch2 compiling that slow is a deal breaker for me. Can you see if [this](https://github.com/catchorg/Catch2/blob/devel/docs/configuration.md#catch_config_fast_compile) helps?

src/api/traits.h
```diff
@@ -0,0 +1,50 @@
...
+#include "hermes_types.h"
+
+namespace hermes {
+namespace api {
+struct BlobInfo {
+  std::string bucket_name;
+  std::string blob_name;
+};
+typedef BlobInfo TraitInput;
+// typedef void* TraitCallback;
```
ChristopherHogan, line 15: Please remove commented out code, or add a comment explaining why it's left in.

src/api/traits.h
```diff
@@ -0,0 +1,50 @@
...
+
+struct Trait;
+typedef std::function<void(TraitInput &, Trait *)> TraitCallback;
+// typedef void(*TraitCallback)(TraitInput &, Trait *);
```
ChristopherHogan, line 19: Please remove commented out code, or add a comment explaining why it's left in.

src/api/traits.h
```diff
@@ -0,0 +1,50 @@
...
+  TraitIdArray conflict_traits;
+  TraitType type;
+  TraitCallback onAttachFn;
+  TraitCallback onDetachFn;
+  TraitCallback onLinkFn;
+  TraitCallback onUnlinkFn;
+  Trait(TraitID id, TraitIdArray conflict_traits, TraitType type);
+};
+
+#define FILE_TRAIT 10
```
ChristopherHogan, line 31: Our convention is to prefix preprocessor symbols with `HERMES_`.

src/api/traits.h
hariharan-devarajan: Done.

src/api/traits.h
hariharan-devarajan: Done

hariharan-devarajan:
> `#include` `traits.cc`



> > The compilation unit for Hermes is the same as you recommended.
> 
> Actually, I recommended that you `#include` `traits.cc` in `buffer_pool.cc` to keep the number of compilation units low, but instead you have `traits.cc` in `HERMES_SRCS` in `src/CMakeLists.txt`, which creates an additional compilation unit. Making that change appears to reduce the time by a few seconds.
> 
> Catch2 compiling that slow is a deal breaker for me. Can you see if [this](https://github.com/catchorg/Catch2/blob/devel/docs/configuration.md#catch_config_fast_compile) helps?

I have added both changes. The reason for keeping trait.cc there was because we have vBucket there as well.

src/api/traits.h
hariharan-devarajan: Done.

src/api/vbucket.h
```diff
@@ -18,20 +18,31 @@ namespace api {
...
+        persist(persist),
+        hermes_(h) {
     LOG(INFO) << "Create VBucket " << initial_name << std::endl;
-    if (hermes_->vbucket_list_.find(initial_name) ==
-        hermes_->vbucket_list_.end())
-      hermes_->vbucket_list_.insert(initial_name);
-    else
-      std::cerr << "VBucket " << initial_name << " exists\n";
+    (void)ctx;
+    if (IsBucketNameTooLong(name_)) {
```
ChristopherHogan, line 41: Shouldn't this be `IsVBucketNameTooLong`?.

src/metadata_management.cc
```diff
@@ -321,15 +326,31 @@ VBucketID GetNextFreeVBucketId(SharedMemoryContext *context, RpcContext *rpc,
...
 
+VBucketID GetOrCreateVBucketId(SharedMemoryContext *context, RpcContext *rpc,
+                               const std::string &name) {
+  MetadataManager *mdm = GetMetadataManagerFromContext(context);
+
+  BeginTicketMutex(&mdm->vbucket_mutex);
+  VBucketID result = GetVBucketIdByName(context, rpc, name.c_str());
+
+  if (result.as_int != 0) {
+    LOG(INFO) << "Opening Bucket '" << name << "'" << std::endl;
```
ChristopherHogan, line 344: "Opening VBucket"

src/metadata_management.cc
```diff
@@ -321,15 +326,31 @@ VBucketID GetNextFreeVBucketId(SharedMemoryContext *context, RpcContext *rpc,
...
+    IncrementRefcount(context, rpc, result);
+  } else {
+    LOG(INFO) << "Creating Bucket '" << name << "'" << std::endl;
```
ChristopherHogan, line 347: "Creating VBucket"

test/trait_test.cc
```diff
@@ -0,0 +1,92 @@
...
+      file_bucket.Put(std::to_string(i), info.write_blob, ctx);
+      blob_names.insert(std::to_string(i));
+    }
+    for (const auto &blob_name : blob_names) {
+      file_vbucket.Link(blob_name, args.filename, ctx);
+      offset_map.emplace(blob_name, std::stol(blob_name) * info.FILE_PAGE);
+    }
+    auto trait =
+        hermes::api::FileMappingTrait(fullpath_str, offset_map, NULL, NULL);
+    file_vbucket.Attach(&trait, ctx);
```
ChristopherHogan, line 86: Shouldn't you open the file you created and `REQUIRE` that the data is what you expect? You're exercising code, but without any assertions we don't know if it's doing the correct thing.

ChristopherHogan:
Unfortunately `CATCH_CONFIG_FAST_COMPILE` doesn't appear to make any difference.

test/bucket_test.cc
```diff
@@ -11,9 +11,16 @@
 #include "test_utils.h"
 
 namespace hapi = hermes::api;
+std::shared_ptr<hermes::api::Hermes> hermes_app;
```
ChristopherHogan, line 14: Why make this a global variable?

hariharan-devarajan: As the callback defined for compression needs to get the blob.

src/metadata_management.cc
hariharan-devarajan: FIXED

src/metadata_management.cc
hariharan-devarajan: FIXED

test/trait_test.cc
hariharan-devarajan: Added the assertions.

src/api/vbucket.h
hariharan-devarajan: FIXED

src/api/vbucket.cc
```diff
@@ -12,12 +16,26 @@ Status VBucket::Link(std::string blob_name, std::string bucket_name,
...
-  LOG(INFO) << "Linking blob "<< blob_name << " in bucket "
-            << bucket_name << " to VBucket " << name_ << '\n';
-
-  // inserting value by insert function
-  linked_blobs_.push_back(make_pair(bucket_name, blob_name));
-
+  LOG(INFO) << "Linking blob " << blob_name << " in bucket " << bucket_name
+            << " to VBucket " << name_ << '\n';
+
+  // FIXME(hari): to use both bucket name and blob name
```
ChristopherHogan, line 22: Is this `FIXME` still relevant since you implemented `BucketContainsBlob`?

hariharan-devarajan: No it is not needed.

ChristopherHogan:
Thanks for the changes. I'll merge this so you can make progress on the adapter, but I think we should discuss alternatives to Catch2. I'm not willing to triple our compilation times.

ChristopherHogan (approved):

hariharan-devarajan:
> Thanks for the changes. I'll merge this so you can make progress on the adapter, but I think we should discuss alternatives to Catch2. I'm not willing to triple our compilation times.

Sure, we can search more if we can optimize this. Or need to switch libraries.

PR #68, opened 2021-01-19
Hariharan/posix adapter

hariharan-devarajan:
This is the adapter implementation PR for STDIO adapter. I have thoroughly created tests for STDIO single process and multi-process (using MPI). I have tested the single process with and without the adapter. Once this PR is accepted I will continue to multi-process with adapter testing.


.Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+ENV PROJECT=$LOCAL/hermes
+ENV INSTALL_DIR=$LOCAL/install
+ENV SPACK_DIR=$LOCAL/spack
+ENV SDS_DIR=$LOCAL/sds
+
+RUN mkdir -p $INSTALL_DIR
+
+RUN git clone https://github.com/spack/spack ${SPACK_DIR}
+RUN git clone https://xgitlab.cels.anl.gov/sds/sds-repo.git ${SDS_DIR}
+RUN git clone --branch hariharan/posix_adapter https://github.com/hariharan-devarajan/hermes ${PROJECT}
```
ChristopherHogan, line 31: If you want to add a `.Dockerfile` could you please set it to use the master branch from the main repo instead of a branch from your fork?

hariharan-devarajan: Done.

adapter/src/hermes/adapter/constants.h
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * Constants file for all adapter.
+ * This file contains common constants across different adapters.
+ */
+
+/**
+ * HERMES_CONF env variable is used to define path to HERMES_CONF in adapters.
+ * This is used for initialization of Hermes.
+ */
+const char* HERMES_CONF = "HERMES_CONF";
```
ChristopherHogan, line 13: Could you follow the Google style for [constant names](https://google.github.io/styleguide/cppguide.html#Constant_Names)?

hariharan-devarajan: Done.

adapter/src/hermes/adapter/interceptor.h
```diff
@@ -0,0 +1,151 @@
...
+/**
+ * HERMES_PRELOAD variable defines if adapter layer should intercept.
+ */
+#ifdef HERMES_PRELOAD
+/**
+ * Typedef and function declaration for intercepted functions.
+ */
+#define HERMES_FORWARD_DECL(__func, __ret, __args) \
+  typedef __ret(*__real_t_##__func) __args;        \
+  __ret(*__real_##__func) __args = NULL;
```
ChristopherHogan, line 99: According to [cppreference](https://en.cppreference.com/w/cpp/language/identifiers), identifiers containing double underscores are reserved.

hariharan-devarajan: Should I use single _ or u have a preference.

ChristopherHogan: I would suggest a single trailing underscore since I use that pattern for private identifiers elsewhere.

hariharan-devarajan: I followed darshan library for the interception style.

hariharan-devarajan: Done. I used trailing single underscore instead.

adapter/src/hermes/adapter/interceptor.h
```diff
@@ -0,0 +1,151 @@
...
+/**
+ * Typedef and function declaration for intercepted functions.
+ */
+#define HERMES_FORWARD_DECL(func_, ret_, args_) \
+  typedef ret_(*real_t_##func_##_) args_;       \
+  ret_(*real_##func_##_) args_ = NULL;
+
+#define HERMES_DECL(func_) func_
+/**
+ * The input function is remained as __real_<func_name>. And a ptr to function
```
ChristopherHogan, line 103: Did you mean "renamed?"

adapter/src/hermes/adapter/interceptor.h
```diff
@@ -0,0 +1,151 @@
...
+ * is obtained using dlsym.
+ */
+#define MAP_OR_FAIL(func_)                                                  \
+  if (!(real_##func_##_)) {                                                 \
+    real_##func_##_ = (real_t_##func_##_)dlsym(RTLD_NEXT, #func_);          \
+    if (!(real_##func_##_)) {                                               \
+      fprintf(stderr, "HERMES Adapter failed to map symbol: %s\n", #func_); \
```
ChristopherHogan, line 110: Shouldn't we use `GLOG` for logging?

adapter/src/hermes/adapter/interceptor.h
hariharan-devarajan: Yes. Thanks. changed.

adapter/src/hermes/adapter/interceptor.h
hariharan-devarajan: I wasnt sure when i wrote this. I changed it.

adapter/src/hermes/adapter/constants.h
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * Constants file for all adapter.
+ * This file contains common constants across different adapters.
+ */
+
+/**
+ * kHermesConf env variable is used to define path to kHermesConf in adapters.
+ * This is used for initialization of Hermes.
+ */
+const char* kHermesConf = "kHermesConf";
```
ChristopherHogan, line 13: Sorry, it makes sense for the environment variable to be named `HERMES_CONF` since that's standard for environment variables. I just meant the identifier should follow Google's conventions.

hariharan-devarajan: I see fixed it as well.

adapter/src/hermes/adapter/interceptor.cc
```diff
@@ -0,0 +1,95 @@
...
+hermes::adapter::InterceptorList list;
+
+namespace hermes::adapter {
+/**
+ * is exit by program called.
+ */
+bool exit = false;
+}  // namespace hermes::adapter
+
+void PopulateBufferingPath() {
```
ChristopherHogan, line 18: Is there a reason the functions in this file are not in the `adapter` namespace?

hariharan-devarajan: These were general functions. But i see your point. I moved them to the namespace as well.

adapter/src/hermes/adapter/interceptor.cc
```diff
@@ -0,0 +1,95 @@
...
+  hermes::u8 config_memory[kConfigMemorySize];
+  if (hermes_config && strlen(hermes_config) > 0) {
+    list.hermes_paths_exclusion.push_back(hermes_config);
+    hermes::Arena config_arena = {};
+    hermes::InitArena(&config_arena, kConfigMemorySize, config_memory);
+    hermes::ParseConfig(&config_arena, hermes_config, &config);
+  } else {
+    InitDefaultConfig(&config);
+  }
+  if (config.mount_points->empty()) {
```
ChristopherHogan, line 30: Why is this check needed? If `config.mount_points` looks like this: `["./nvme", "./burst_buffer"]`, then these paths won't get added to your exclusion list, right?

hariharan-devarajan: oops. I intended the not.

adapter/src/hermes/adapter/interceptor.cc
```diff
@@ -0,0 +1,95 @@
...
+  hermes::u8 config_memory[kConfigMemorySize];
+  if (hermes_config && strlen(hermes_config) > 0) {
+    list.hermes_paths_exclusion.push_back(hermes_config);
+    hermes::Arena config_arena = {};
+    hermes::InitArena(&config_arena, kConfigMemorySize, config_memory);
+    hermes::ParseConfig(&config_arena, hermes_config, &config);
+  } else {
+    InitDefaultConfig(&config);
+  }
+  if (!config.mount_points->empty()) {
```
ChristopherHogan, line 30: I think you just want to drop that conditional altogether. If `config.mount_points` is `["", "./nvme", "./burst_buffer"]` (which is common since the empty string represents RAM), this will fail to add the mount points to your exclusion list. You already filter out the empty string correctly on line 32. Maybe you thought `mount_points` was a `std::vector`? It's just an array of `std::string`.

hariharan-devarajan: I see. ok, i removed the condition. The condition was if the array of strings was empty. Either way the for will take care of it.

adapter/src/hermes/adapter/interceptor.cc
```diff
@@ -0,0 +1,94 @@
...
+    }
+  }
+  return true;
+}
+bool IsTracked(FILE* fh) {
+  if (hermes::adapter::exit) return false;
+  atexit(OnExit);
+  int MAXSIZE = 0xFFF;
+  char proclnk[0xFFF];
+  char filename[0xFFF];
```
ChristopherHogan, line 81: I assume you meant to use your `MAXSIZE` variable instead of repeating the literal?

hariharan-devarajan: Correct. Somehow i got sidetracked :(. Fixed.

src/buffer_pool.cc
```diff
@@ -1355,7 +1355,7 @@ size_t LocalWriteBufferById(SharedMemoryContext *context, BufferID id,
       context->open_streams[device->id][slab_index] = file;
     }
     fseek(file, header->data_offset, SEEK_SET);
-    [[maybe_unused]] size_t items_written = fwrite(at, write_size, 1, file);
```
ChristopherHogan, line 1358: Did your compiler not like this attribute? It's [supposedly](https://gcc.gnu.org/projects/cxx-status.html) supported since GCC 7.

adapter/test/stdio/stdio_adapter_test.cpp
```diff
@@ -0,0 +1,82 @@
...
+  fullpath /= args.filename;
+  info.new_file = fullpath.string() + "_new";
+  info.existing_file = fullpath.string() + "_ext";
+  if (fs::exists(info.new_file)) fs::remove(info.new_file);
+  if (fs::exists(info.existing_file)) fs::remove(info.existing_file);
+  if (!fs::exists(info.existing_file)) {
+    std::string cmd = "dd if=/dev/zero of=" + info.existing_file + " bs=" +
+                      std::to_string(args.request_size * info.num_iterations) +
+                      " count=1  > /dev/null 2>&1";
+    system(cmd.c_str());
```
ChristopherHogan, line 56: On gcc 9.3 I'm getting several `unused-result` warnings everywhere you're calling `system`. Could you `REQUIRE` that the return code is valid.

adapter/test/posix/posix_adapter_rs_test.cpp
```diff
@@ -0,0 +1,1184 @@
...
+ **/
+
+TEST_CASE("BatchedWriteRSRangeLarge",
+          "[process="+std::to_string(info.comm_size)+"]"
+          "[operation=batched_write]"
+          "[request_size=range-large][repetition=1024]"
+          "[pattern=sequential][file=1]") {
+    pretest();
+    SECTION("write to new file always at the start") {
+        int fd = open(info.new_file.c_str(), O_WRONLY | O_CREAT);
```
ChristopherHogan, line 798: I'm getting compilation errors in gcc 9.3 everywhere `open` is called with the `O_CREAT` flag and without a `mode` argument. From `man open(2)`:
> The mode argument must be supplied if O_CREAT or O_TMPFILE is specified in flags; if it is not supplied, some arbitrary bytes from the stack will be applied as the file mode.

Here's an example error message:
```
In function ‘int open(const char*, int, ...)’,
    inlined from ‘void ____C_A_T_C_H____T_E_S_T____206()’ at /home/chogan/dev/hermes/adapter/test/posix/posix_adapter_rs_test.cpp:798:22:
/usr/include/x86_64-linux-gnu/bits/fcntl2.h:50:24: error: call to ‘__open_missing_mode’ declared with attribute error: open with O_CREAT or O_TMPFILE in second argument needs 3 arguments
   50 |    __open_missing_mode ();
      |    ~~~~~~~~~~~~~~~~~~~~^~

```

CMakeLists.txt
```diff
@@ -349,6 +348,7 @@ endif()
 if(CMAKE_PROJECT_NAME STREQUAL HERMES AND BUILD_TESTING)
   find_package(Catch2 REQUIRED)
   add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/test)
+  add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/adapter/test)
```
ChristopherHogan, line 351: Could you guard the building and testing of the adapters with a `HERMES_ENABLE_ADAPTERS` CMake option? The default can be `ON`, but it would be nice to disable the adapters when doing library development.

src/buffer_pool.cc
hariharan-devarajan: yeah. I saw u used it on next line in the assertion. I reverted the change.

adapter/test/stdio/stdio_adapter_test.cpp
hariharan-devarajan: Done. This was the case across all test cases.

adapter/test/posix/posix_adapter_rs_test.cpp
hariharan-devarajan: I m using gcc 8 and so I am not getting these errors. Anyhow, I have added the mode attribute to calls where I use O_CREATE.

CMakeLists.txt
hariharan-devarajan: added option HERMES_ENABLE_ADAPTERS.

adapter/src/hermes/adapter/stdio/metadata_manager.h
```diff
@@ -0,0 +1,115 @@
...
+  /**
+   * Get the instance of hermes.
+   */
+  std::shared_ptr<hapi::Hermes>& GetHermes() { return hermes; }
+
+  /**
+   * Initialize hermes. Get the kHermesConf from environment else get_env
+   * returns NULL which is handled internally by hermes. Initialize hermes in
+   * daemon mode. Keep a reference of how many times Initialize is called.
+   * Within the adapter, Initialize is called from fopen.
```
ChristopherHogan, line 54: Is there a reason you didn't use the `_init` mechanism to initialize Hermes once when the adapter `.so` is loaded?

adapter/src/hermes/adapter/stdio/metadata_manager.h
```diff
@@ -0,0 +1,115 @@
...
+   */
+  std::unordered_map<FileID, AdapterStat> metadata;
+  /**
+   * hermes attribute to initialize Hermes
+   */
+  std::shared_ptr<hapi::Hermes> hermes;
+  /**
+   * references of how many times hermes was tried to initialize.
+   */
+  size_t ref;
```
ChristopherHogan, line 38: Shouldn't this be a `std::atomic<size_t>`?

adapter/src/hermes/adapter/stdio/metadata_manager.h
hariharan-devarajan: U mean MPI init?

adapter/src/hermes/adapter/stdio/metadata_manager.cc
```diff
@@ -0,0 +1,56 @@
...
+  if (iter == metadata.end())
+    return MetadataReturn(AdapterStat(), false);
+  else
+    return MetadataReturn(iter->second, true);
+}
+
+FileID MetadataManager::Convert(FILE *fh) {
+  struct stat st;
+  int fd = fileno(fh);
+  fstat(fd, &st);
```
ChristopherHogan, line 42: Probably should check the return code of `fstat`.

adapter/src/hermes/adapter/stdio/metadata_manager.h
ChristopherHogan: No. The `-init` flag to `ld` lets you specify a function that is called when the shared library loads. That way you won't have to check for Hermes initialization in every `fopen` or keep a refcount.

adapter/test/stdio/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
+function(gcc_hermes exec tag_name tags)
+    add_test(NAME Test${exec}_${tag_name} COMMAND "${CMAKE_BINARY_DIR}/bin/${exec}" ${tags} -s --reporter compact)
+    set_property(TEST Test${exec}_${tag_name}
+            PROPERTY ENVIRONMENT HERMES_CONF=${CMAKE_SOURCE_DIR}/adapter/test/data/hermes.conf)
+    set_property(TEST Test${exec}_${tag_name} APPEND
+            PROPERTY ENVIRONMENT LSAN_OPTIONS=suppressions=${CMAKE_SOURCE_DIR}/test/data/asan.supp)
+endfunction()
+#------------------------------------------------------------------------------
+# Posix Adapter Internal tests
```
ChristopherHogan, line 9: Posix -> Stdio

adapter/test/stdio/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+#------------------------------------------------------------------------------
+add_executable(stdio_adapter_mapper_test stdio_adapter_mapper_test.cpp ${ADAPTER_COMMON})
+target_link_libraries(stdio_adapter_mapper_test
+        hermes_stdio Catch2::Catch2 -lstdc++fs -ldl -lc MPI::MPI_CXX)
+add_dependencies(stdio_adapter_mapper_test hermes_stdio)
+
+#------------------------------------------------------------------------------
+# Posix Adapter End to End tests
```
ChristopherHogan, line 17: Posix -> Stdio

adapter/test/posix/posix_adapter_rs_test.cpp
ChristopherHogan: I'm still getting compilation errors. You also need to add the mode when you use `O_TMPFILE`.

adapter/src/hermes/adapter/stdio/metadata_manager.h
hariharan-devarajan: Do u have a reference i can follow.

adapter/test/stdio/CMakeLists.txt
hariharan-devarajan: Done

adapter/src/hermes/adapter/stdio/metadata_manager.cc
hariharan-devarajan: Done

adapter/test/stdio/CMakeLists.txt
hariharan-devarajan: Done

adapter/src/hermes/adapter/stdio/metadata_manager.h
hariharan-devarajan: Done

adapter/src/hermes/adapter/stdio/metadata_manager.h
ChristopherHogan: There's a descent summary [here](https://stackoverflow.com/a/9759936/2059743). It looks like `__attribute__(constructor)` is preferable to `ld -init` these days. This is a GCC extension though, so if people are using other compilers on the machines we're targeting, maybe we should leave it as is.

hariharan-devarajan: Yes that is what i read. I will explore this option and let u know.,

hariharan-devarajan: I0121 12:15:49.370560  5749 memory_management.cc:128] PushSize added 2 bytes of padding for alignment
I0121 12:15:49.373865  5749 memory_management.cc:128] PushSize added 2 bytes of padding for alignment
Attempting to use an MPI routine before initializing MPICH

The constructor is called before main so MPI is not initialized. I think this approach won't work for us. We should do the init on fopen.

adapter/test/stdio/stdio_adapter_basic_test.cpp
```diff
@@ -0,0 +1,1293 @@
...
+    int status = fclose(fd);
+    REQUIRE(status == 0);
+    fd = fopen(info.existing_file.c_str(), "r");
+    REQUIRE(fd != nullptr);
+    status = fclose(fd);
+    REQUIRE(status == 0);
+  }
+
+  SECTION("append write new file") {
+    FILE* fd = fopen(info.existing_file.c_str(), "a");
```
ChristopherHogan, line 38: The `SECTION` says "new file" but you're opening `existing_file`.

adapter/test/stdio/stdio_adapter_basic_test.cpp
```diff
@@ -0,0 +1,1293 @@
...
+    REQUIRE(fd != nullptr);
+    int status = fclose(fd);
+    REQUIRE(status == 0);
+  }
+
+  SECTION("append write and read new file") {
+    FILE* fd = fopen(info.existing_file.c_str(), "a+");
```
ChristopherHogan, line 45: The `SECTION` says "new file" but you're opening `existing_file`.

adapter/test/posix/posix_adapter_rs_test.cpp
hariharan-devarajan: Done.

adapter/test/stdio/stdio_adapter_basic_test.cpp
hariharan-devarajan: fixed

adapter/test/stdio/stdio_adapter_basic_test.cpp
hariharan-devarajan: fixed

adapter/test/posix/posix_adapter_rs_test.cpp
ChristopherHogan: There's still one more error. In `posix_adapter_basic_test.cpp:1329` you're passing 4 arguments to `open`.

hariharan-devarajan: My compiler missed it. Fixed it now.

adapter/src/hermes/adapter/interceptor.cc
```diff
@@ -0,0 +1,94 @@
+/**
+ * Internal headers
+ */
+#include "interceptor.h"
+
+/**
+ * Define Interceptor list for adapter.
+ */
+hermes::adapter::InterceptorList list;
```
ChristopherHogan, line 9: `hermes_stdio_adapter_test` is failing on my system (Ubuntu 20.04). Before `main`, ` libnuma.so.1` attempts to call `fopen` on `/proc/self/status` which the adapter intercepts. However, `IsTracked()` returns `true` for this file because `hermes::adapter::InterceptorList list` has been zero-initialized by the compiler (because it's storage duration is static) but the default constructor has not been called yet (i.e., its `path_exclusions` member is empty). So, the adapter attempts to buffer the file but Hermes initialization fails because MPI isn't initialized yet. I guess we need to create a default `InterceptorList` on the stack if the static `list` isn't initialized yet (which we can presumably detect if `path_exclusions` is empty, though I'm not sure how safe it is to access `std::` data structures that are zero-initialized). Thoughts?

ChristopherHogan: > though I'm not sure how safe it is to access std:: data structures that are zero-initialized

According to the "Access Outside of Lifetime" section [here](https://en.cppreference.com/w/cpp/language/lifetime), it's undefined behavior.

hariharan-devarajan: Ok, what about moving path exclusion and inclusions as constant in interceptor.h and then using them as lists directly. We support user_path exclusions anyway which can be used by users.

ChristopherHogan: Still the same problem. `kPathExclusions` is empty the first time `IsTracked` is called. Can you somehow eliminate the global `list`?

ChristopherHogan: It works if you switch `kPathExclusions` from a vector to a `char*` array:
```c++
const char *kPathExclusions[] = {
    "/bin/",  "/boot/", "/dev/", "/etc/", "/lib/", "/opt/",
    "/proc/", "/sbin/", "/sys/", "/usr/", "/var/"
};
```
Although we're still relying on undefined behavior because we call `list.push_back` before its constructor is called. Best would be to eliminate the global.

ChristopherHogan: Also, can you please add `"/run/"` to the exclusions list?

hariharan-devarajan: So you want me to switch to ```const char *kPathExclusions[]``` and add /run/. Is that correct?

hariharan-devarajan: Also to remove the global list, I can create a singleton. I have tested it and commit the changes. Let me know if that approach works for you.

ChristopherHogan: The singleton seems to work. As a side note, initializing and finalizing Hermes several thousand times in a single process has reminded me that I never close the buffering files. I can do that in a separate PR though.

adapter/src/hermes/adapter/interceptor.cc
```diff
@@ -0,0 +1,95 @@
+/**
+ * Internal headers
+ */
+#include "interceptor.h"
+
+/**
+ * Define Interceptor list for adapter.
+ */
+#define LIST \
```
ChristopherHogan, line 9: Could you use a more descriptive name since this token is globally visible?

adapter/src/hermes/adapter/interceptor.cc
hariharan-devarajan: Ok sounds good

adapter/src/hermes/adapter/interceptor.cc
hariharan-devarajan: Done.

hariharan-devarajan:
I have added the correctness test in the trait test for checking if the order of data is correct.

adapter/test/stdio/stdio_adapter_basic_test.cpp
```diff
@@ -0,0 +1,1293 @@
...
+  }
+  posttest();
+}
+
+TEST_CASE("BatchedReadRandomRSVariable",
+          "[process=" + std::to_string(info.comm_size) +
+              "]"
+              "[operation=batched_read]"
+              "[request_size=type-variable]"
+              "[repetition=1024][pattern=random][file=1]") {
```
ChristopherHogan, line 512: Should `repetition` be bound to `info.num_iterations`, or does the 1024 come from somewhere else?

hariharan-devarajan: added it as a variable to tags.

adapter/test/stdio/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+            PROPERTY ENVIRONMENT LSAN_OPTIONS=suppressions=${CMAKE_SOURCE_DIR}/test/data/asan.supp)
+endfunction()
+#------------------------------------------------------------------------------
+# STDIO Adapter Internal tests
+#------------------------------------------------------------------------------
+add_executable(stdio_adapter_mapper_test stdio_adapter_mapper_test.cpp ${ADAPTER_COMMON})
+target_link_libraries(stdio_adapter_mapper_test
+        hermes_stdio Catch2::Catch2 -lstdc++fs -ldl -lc MPI::MPI_CXX)
+add_dependencies(stdio_adapter_mapper_test hermes_stdio)
+
```
ChristopherHogan, line 15: I think you're missing a `gcc(stdio_adapter_mapper_test "")` here. The mapper test isn't being run in `ctest` (unless I'm missing something).

hariharan-devarajan: Added.

adapter/src/hermes/adapter/stdio/mapper/balanced_mapper.cc
```diff
@@ -0,0 +1,33 @@
...
+using hermes::adapter::stdio::HermesStruct;
+
+MapperReturnType BalancedMapper::map(const FileStruct& file_op) {
+  LOG(INFO) << "Mapping File with offset:" << file_op.offset_
+            << " and size:" << file_op.size_ << "." << std::endl;
+
+  auto mapper_return = MapperReturnType();
+  size_t size_mapped = 0;
+  while (file_op.size_ > size_mapped) {
+    FileStruct file;
```
ChristopherHogan, line 17: Do you need to copy `file_op.file_id_` into `file.file_id_` here? Every mapping returned from this function contains the default `FileID` of `{0,0}`. Is that a bug or intentional?

hariharan-devarajan: It doesn't have any impact on the logic. But your right, I will add it either way.

hariharan-devarajan:
I will recreate this pull request. I will pulling the changes into my mpi branch somehow it is showing here. :(

hariharan-devarajan:
Sorry for the confusion. I have reverted the merge.

adapter/src/hermes/adapter/stdio/mapper/balanced_mapper.cc
ChristopherHogan: If the logic doesn't depend on the `FileID`, then how do you detect if two processes open the same file?

ChristopherHogan:
The merge fiasco seems to have undone the following commits: https://github.com/HDFGroup/hermes/pull/68/commits/58198454ac94696fe0ac07295c17f119a99e20cd, https://github.com/HDFGroup/hermes/pull/68/commits/7bd4aa79b2a24200faa128db60c08c04ea41bed9, https://github.com/HDFGroup/hermes/pull/68/commits/1a14c314d1f8fcddb0fb09588c90257c6fd696ed. There might be more. Please verify by making sure that `HEAD` is the same as the last known good commit, i.e.,: `git diff 1a14c31 HEAD`.

hariharan-devarajan:
> The merge fiasco seems to have undone the following commits: [5819845](https://github.com/HDFGroup/hermes/commit/58198454ac94696fe0ac07295c17f119a99e20cd), [7bd4aa7](https://github.com/HDFGroup/hermes/commit/7bd4aa79b2a24200faa128db60c08c04ea41bed9), [1a14c31](https://github.com/HDFGroup/hermes/commit/1a14c314d1f8fcddb0fb09588c90257c6fd696ed). There might be more. Please verify by making sure that `HEAD` is the same as the last known good commit, i.e.,: `git diff 1a14c31 HEAD`.

I have reverted the change. during reverting, I fixed few other issues as well. 
- CMakeLists of stdio test the stdio_adapter_test was getting called along with hermes. 
- I fixed the tags for stdio_adapter_func_test.

adapter/src/hermes/adapter/stdio/mapper/balanced_mapper.cc
hariharan-devarajan: That is outside of this. It is done during open. And a respective bucket is opened/created. At this stage, the mapper just maps offset and size to blobs.

ChristopherHogan: Ok.

adapter/src/hermes/adapter/stdio/stdio.cc
```diff
@@ -0,0 +1,993 @@
...
+  MAP_OR_FAIL(freopen);
+  auto mdm = hermes::adapter::Singleton<MetadataManager>::GetInstance();
+  ret = real_freopen_(path_str.c_str(), mode, stream);
+  if (!ret) {
+    return ret;
+  } else {
+    auto existing = mdm->Find(ret);
+    if (!existing.second) {
+      LOG(INFO) << "Reopen file for filename " << path_str << " in mode "
+                << mode << std::endl;
```
ChristopherHogan, line 88: Shouldn't this log message be outside of the current `if` block?

adapter/src/hermes/adapter/stdio/stdio.cc
```diff
@@ -0,0 +1,993 @@
...
+  for (const auto &item : mapping) {
+    hapi::Context ctx;
+    auto blob_exists =
+        existing.first.st_bkid->ContainsBlob(item.second.blob_name_);
+    hapi::Blob put_data((unsigned char *)ptr + data_offset,
+                        (unsigned char *)ptr + data_offset + item.first.size_);
+    existing.first.st_blobs.emplace(item.second.blob_name_);
+    if (!blob_exists || item.second.size_ == kPageSize) {
+      LOG(INFO) << "Overwrite blob " << item.second.blob_name_
+                << " of size:" << item.second.size_ << "." << std::endl;
```
ChristopherHogan, line 127: We're not overwriting the blob if it doesn't exist, right?

adapter/src/hermes/adapter/stdio/stdio.cc
```diff
@@ -0,0 +1,993 @@
...
+      existing.first.st_bkid->Put(item.second.blob_name_, put_data, ctx);
+    } else {
+      LOG(INFO) << "Blob " << item.second.blob_name_
+                << " of size:" << item.second.size_ << " exists." << std::endl;
+      hapi::Blob temp(0);
+      auto exiting_blob_size =
```
ChristopherHogan, line 133: did you mean "existing?"

hariharan-devarajan: Yes. Done.

adapter/src/hermes/adapter/stdio/stdio.cc
hariharan-devarajan: It has both. Create or Overrite. I updated the comment.

adapter/src/hermes/adapter/stdio/stdio.cc
hariharan-devarajan: Yap fixed.

adapter/src/hermes/adapter/stdio/stdio.cc
```diff
@@ -0,0 +1,993 @@
...
+    }
+    status = fclose(fh);
+  }
+  INTERCEPTOR_LIST->hermes_flush_exclusion.erase(filename);
+  return read_size;
+}
+
+size_t read_internal(std::pair<AdapterStat, bool> &existing, void *ptr,
+                     size_t total_size, FILE *fp) {
+  LOG(INFO) << "Write called for filename: "
```
ChristopherHogan, line 240: Shouldn't this be "Read called?"

hariharan-devarajan: Yes. Done

adapter/src/hermes/adapter/stdio/stdio.cc
```diff
@@ -0,0 +1,992 @@
...
+FILE *HERMES_DECL(fopen64)(const char *path, const char *mode) {
+  FILE *ret;
+  std::string path_str(path);
+  if (hermes::adapter::IsTracked(path)) {
+    LOG(INFO) << "Intercept fopen64 for filename: " << path
+              << " and mode: " << mode << " is tracked." << std::endl;
+    ret = open_internal(path, mode);
+  } else {
+    MAP_OR_FAIL(fopen64);
+    ret = real_fopen_(path, mode);
```
ChristopherHogan, line 351: `real_fopen64_`?

adapter/src/hermes/adapter/stdio/stdio.cc
```diff
@@ -0,0 +1,992 @@
...
+FILE *HERMES_DECL(fdopen)(int fd, const char *mode) {
+  FILE *ret;
+  MAP_OR_FAIL(fdopen);
+  ret = real_fdopen_(fd, mode);
+  if (ret && hermes::adapter::IsTracked(ret)) {
+    LOG(INFO) << "Intercept fdopen for file descriptor: " << fd
+              << " and mode: " << mode << " tracked." << std::endl;
+    int MAXSIZE = 0xFFF;
+    char proclnk[0xFFF];
+    char filename[0xFFF];
```
ChristopherHogan, line 365: Use `MAXSIZE` instead of repeating literals.

hariharan-devarajan: Fixed

adapter/src/hermes/adapter/stdio/stdio.cc
hariharan-devarajan: Fixed

adapter/src/hermes/adapter/stdio/stdio.cc
```diff
@@ -0,0 +1,992 @@
...
+            offset_map.emplace(blob_name, std::stol(blob_name) * kPageSize);
+          }
+          auto trait = hermes::api::FileMappingTrait(filename, offset_map,
+                                                     nullptr, NULL, NULL);
+          file_vbucket.Attach(&trait, ctx);
+          file_vbucket.Delete(ctx);
+          existing.first.st_blobs.clear();
+          INTERCEPTOR_LIST->hermes_flush_exclusion.erase(filename);
+        }
+        existing.first.st_bkid->Close(ctx);
```
ChristopherHogan, line 469: I think here you want `existing.first.st_bkid->Destroy(ctx);` `Close` decrements the internal refcount but doesn't free any resources. You should call `Close` in the `else` block a few lines down.

adapter/src/hermes/adapter/stdio/stdio.cc
```diff
@@ -0,0 +1,992 @@
...
+        mdm->FinalizeHermes();
```
ChristopherHogan, line 470: I know why you're doing this, but it seems inefficient to shutdown Hermes on every `fclose`. We should try to think of a better way as we go forward.

hariharan-devarajan: It won't shut down in MPI case. If it is a pure STDIO code we can initialize and finalize only at open and close respectively.

adapter/src/hermes/adapter/stdio/stdio.cc
hariharan-devarajan: Ok fixed.

adapter/src/hermes/adapter/stdio/stdio.cc
ChristopherHogan: But if a `stdio` program writes 1,000 files for example, we do tons of unnecessary work, right?

hariharan-devarajan: I agree. However, we don't have any other init and finalize part. For the MPI case I do intercept MPI init and finalize which would be the case for most hpc cases.

hariharan-devarajan: Plus if users first open multiple times it will only initialize once and finalize at last close. Only if they open and close per file one by one it's bad .

adapter/src/hermes/adapter/stdio/stdio.cc
```diff
@@ -0,0 +1,993 @@
...
+}
+
+int HERMES_DECL(getw)(FILE *stream) {
+  int ret;
+  if (hermes::adapter::IsTracked(stream)) {
+    auto mdm = hermes::adapter::Singleton<MetadataManager>::GetInstance();
+    auto existing = mdm->Find(stream);
+    if (existing.second) {
+      LOG(INFO) << "Intercept getw." << std::endl;
+      unsigned char value;
```
ChristopherHogan, line 724: Isn't `getw` supposed to read an `int`?

hariharan-devarajan: With int the char conversion fails with undefined behavior. Unsigned char gives most consistent results.

ChristopherHogan: But you're only reading 1 byte. It's supposed to read `sizeof(int)` bytes.

adapter/src/hermes/adapter/stdio/stdio.cc
```diff
@@ -0,0 +1,993 @@
...
+}
+
+char *HERMES_DECL(fgets)(char *s, int size, FILE *stream) {
+  char *ret;
+  if (hermes::adapter::IsTracked(stream)) {
+    auto mdm = hermes::adapter::Singleton<MetadataManager>::GetInstance();
+    auto existing = mdm->Find(stream);
+    if (existing.second) {
+      LOG(INFO) << "Intercept fgets." << std::endl;
+      read_internal(existing, s, size, stream);
```
ChristopherHogan, line 745: I don't think this preserves the semantics of [fgets](https://linux.die.net/man/3/fgets):

> fgets() reads in at most one less than size characters from stream and stores them into the buffer pointed to by s. Reading stops after an EOF or a newline. If a newline is read, it is stored into the buffer. A terminating null byte (aq\0aq) is stored after the last character in the buffer.

hariharan-devarajan: So i need to do the following.
Read size - 1 data
if I find \0 or \n before size - 1 i should stop and return that buffer. Correct?

adapter/src/hermes/adapter/stdio/stdio.cc
hariharan-devarajan: Ok, i tried that. It worked.

adapter/src/hermes/adapter/stdio/stdio.cc
ChristopherHogan: Yes, and include a null terminating byte.

hariharan-devarajan: Ok fixed.

adapter/src/hermes/adapter/stdio/stdio.cc
```diff
@@ -631,9 +631,9 @@ int HERMES_DECL(fgetc)(FILE *stream) {
     auto existing = mdm->Find(stream);
     if (existing.second) {
       LOG(INFO) << "Intercept fgetc." << std::endl;
-      unsigned char value;
-      auto ret_size = read_internal(existing, &value, sizeof(value), stream);
-      if (ret_size == 1) {
+      int value;
```
ChristopherHogan, line 634: I think `fgetc`, `getc`, and `_IO_getc` were correct before in that they are only supposed to read one byte. It is only `getw` that reads `sizeof(int)` bytes.

hariharan-devarajan: Reverted the changes for fgetc, getc and _IO_getc

adapter/src/hermes/adapter/stdio/stdio.cc
```diff
@@ -0,0 +1,1017 @@
...
+}
+
+int HERMES_DECL(fgetpos)(FILE *fp, fpos_t *pos) {
+  int ret;
+  if (hermes::adapter::IsTracked(fp) && pos) {
+    auto mdm = hermes::adapter::Singleton<MetadataManager>::GetInstance();
+    auto existing = mdm->Find(fp);
+    if (existing.second) {
+      LOG(INFO) << "Intercept fgetpos." << std::endl;
+      pos->__pos = existing.first.st_ptr;
```
ChristopherHogan, line 537: Could you please add the following comment in the 3 functions (`fgetpos`, `fsetpos`, and `fsetpos64`) where you access `fpos_t->__pos`:

```c++
// TODO(chogan): @portability In the GNU C Library, fpos_t is an opaque data
// structure that contains internal data to represent file offset and conversion
// state information. In other systems, it might have a different internal
// representation. This will need to change to support other compilers.

```

hariharan-devarajan: Added

ChristopherHogan (approved):
Thanks for all the changes. I'll go ahead and merge this, but I have a short wish list that you will hopefully keep in mind as you continue working on the adapter.

1. It would be nice if the adapter code were included in our [CI coverage reports](https://coveralls.io/github/HDFGroup/hermes)
2. There is a lot of duplicated code under `adapter/test`. Hopefully as you add functionality you can do some refactoring to reduce the duplication.
3. There is no data verification in the tests. We test that `fread` and `fwrite` read or write the expected number of bytes, but we never verify that the read/written data is what we expect. In theory the adapter could be reading and writing random bytes and the tests would never catch that. I think a more thourough test looks like this:

```

fopen()
fwrite()
fclose()
fsync()

The data has now gone from the application, through the adapter, into Hermes,
and into the filesystem. Now we have to make sure the data that ended up on
the filesystem is correct.

real_fopen_()
real_fread_()
verify that the contents of the file produced by the adapter is what we expect.
real_fclose_()
```

In my opinion, the current tests are wasteful and not useful because they read and write the same uniform data thousands of times without any verification. Varying the offset, stride and size doesn't accomplish anything if you're not verifying the data.

PR #113, opened 2021-02-02
Kimmy/copying

jya-kmu:
Fix alignment issue.


ChristopherHogan:
Several files (`adapter/src/hermes/adapter/stdio/common/constants.h` for example) still reference "COPYFILE" instead of "the COPYING file."

ChristopherHogan (approved):

PR #115, opened 2021-02-04
Bug fixes for Ares

ChristopherHogan:
* Hermes cores need RPC clients. Previously just the application cores had them.
* Converted several RPCs that returned `void` to return a `bool`. We don't actually use the result, but it has the effect of making the call synchronous. Previously, when the calls were asynchronous (meaning the caller didn't have to wait for a response), we had an eventual consistency model, so it was possible for a `Get` to not see metadata that an immediately preceding `Put` updated. We may switch back to eventual consistency at some point for performance reasons, but right now I prefer deterministic correctness.

### DIFF
--- a/src/api/bucket.cc
+++ b/src/api/bucket.cc
@@ -87,8 +87,10 @@ size_t Bucket::GetBlobSize(Arena *arena, const std::string &name,
               << name_ << '\n';
     BlobID blob_id = GetBlobIdByName(&hermes_->context_, &hermes_->rpc_,
                                      name.c_str());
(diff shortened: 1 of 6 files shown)


jya-kmu:
If it can simplify something at this point it is fine. Yet we may create a bookmark for this change since we expect performance issue in the future especially when scaling.

jya-kmu (approved):
If it can simplify something at this point it is fine. Yet we may create a bookmark for this change since we expect performance issue in the future especially when scaling.

PR #136, opened 2021-02-17
Kimmy/dpe benchmark

jya-kmu:


src/utils.h
```diff
@@ -93,5 +94,88 @@ size_t RoundDownToMultiple(size_t val, size_t multiple);
...
+  std::multimap<hermes::u64, TargetID> ordered_cap;
+  int num_devices;
+};
+
+/**
+ * Initialize device state with default values.
+ *
+ * @param total_target Total number of target.
+ *
+ * @param even_dist The device distribution is homogeneous or not.
```
ChristopherHogan, line 120: The docs say `even_dist` but the argument is named `homo_dist`.

src/utils.cc
```diff
@@ -106,4 +111,151 @@ void InitDefaultConfig(Config *config) {
   }
   config->buffer_pool_shmem_name[shmem_name_size] = '\0';
 }
+
+namespace testing {
+
+/** Use Megabytes */
+std::vector<i64> device_size {128, 1024, 4096, 16384};
```
ChristopherHogan, line 118: Since this is only used in one function (`InitDeviceState`) can you define it in that function instead of using a global variable?

src/utils.cc
```diff
@@ -106,4 +111,151 @@ void InitDefaultConfig(Config *config) {
...
+
+/** Use Megabytes/Sec */
+std::vector<double> device_bandwidth {8192, 3072, 550, 120};
```
ChristopherHogan, line 121: Same comment as `device_size`.

src/utils.cc
```diff
@@ -106,4 +111,151 @@ void InitDefaultConfig(Config *config) {
...
+
+std::vector<double> tgt_homo_dist {0.25, 0.25, 0.25, 0.25};
+std::vector<double> tgt_heto_dist {0.1, 0.2, 0.3, 0.4};
```
ChristopherHogan, line 124: Same comment as `device_size`.

src/utils.cc
```diff
@@ -106,4 +111,151 @@ void InitDefaultConfig(Config *config) {
...
+
+std::vector<size_t> GenFixedNumberOfBlobs(int num,
+                                          size_t each_blob_size) {
+  std::vector<size_t> result;
+
+  for (auto i {0}; i < num; ++i) {
+    result.push_back(each_blob_size);
+  }
+  return result;
+}
```
ChristopherHogan, line 258: This functionality can be achieved more efficiently (in terms of amount of code and number of instructions) with the `std::vector` constructor that takes a size and a default value: `std::vector<size_t> vec(num, each_blob_size);`. I would just replace all calls to `GetFixedNumberOfBlobs(num, each_blob_size);` with `std::vector<T>(num, each_blob_size)`.

benchmarks/dpe_bench.cc
```diff
@@ -0,0 +1,189 @@
...
+#include "utils.h"
+#include "data_placement_engine.h"
+
+/* example usage: ./bin/dpe_bench -m -s 4096 */
+
+using namespace hermes;  // NOLINT(*)
+using std::chrono::time_point;
+const auto now = std::chrono::high_resolution_clock::now;
+
+enum class PlacementPolicy {
```
ChristopherHogan, line 16: Can't you use `hermes::api::PlacementPolicy` instead of redefining the policies here?

benchmarks/dpe_bench.cc
```diff
@@ -0,0 +1,189 @@
...
+  kRandom,
+  kRoundRobin,
+  kMinimizeIoTime,
+};
+
+#define TOTAL_TARGETS 10
+#define BLOB_RANGE testing::BlobSizeRange::kHuge
+
+#define TOTAL_NUM_BLOBS 10
+#define TOTAL_BLOB_SIZE GIGABYTES(10)
```
ChristopherHogan, line 26: The [Google style](https://google.github.io/styleguide/cppguide.html#Preprocessor_Macros) suggests preferring `const` variables over macros.

benchmarks/dpe_bench.cc
```diff
@@ -0,0 +1,189 @@
...
+    blob_sizes = testing::GenFixedNumberOfBlobs(TOTAL_NUM_BLOBS,
+                                                each_blob_size);
+    total_placed_size = each_blob_size * TOTAL_NUM_BLOBS;
+  }
+
+  testing::TargetViewState tgt_state = testing::InitDeviceState(TOTAL_TARGETS);
+
+  assert(tgt_state.num_targets == TOTAL_TARGETS);
+  std::cout << "Initialize Device State.\n";
+  /* testing::PrintNodeState(tgt_state); */
```
ChristopherHogan, line 126: Is this debug code that can be removed?

benchmarks/dpe_bench.cc
```diff
@@ -0,0 +1,189 @@
...
+  if (!result) {
+    for (auto it = output_tmp.begin(); it != output_tmp.end(); ++it) {
+      PlacementSchema schema = AggregateBlobSchema((*it));
+      assert(schema.size() > 0);
+      schemas.push_back(schema);
+    }
+  }
+
+  std::cout << "Blob placement is done.\n";
+  /* testing::PrintNodeState(tgt_state); */
```
ChristopherHogan, line 184: Is this debug code that can be removed?

src/utils.cc
jya-kmu: Right, this function is silly...

ChristopherHogan (approved):

PR #139, opened 2021-02-23
Kimmy/error handling

jya-kmu:
Open for discussion.


ChristopherHogan:
Github is complaining about some conflicting files.

jya-kmu:
> Github is complaining about some conflicting files.

Yes, will be on this after done with hdf5 doxygen.

src/hermes_types.h
```diff
@@ -42,7 +42,6 @@ typedef double f64;
 typedef u16 DeviceID;
 
 namespace api {
-typedef int Status;
```
ChristopherHogan, line 45: Is there a reason you moved the `Status` from the `hermes::api` namespace to the `hermes` namespace? In my opinion, every object that a Hermes-native user will interact with belongs in `hermes::api` and `src/api`.

src/hermes_status.h
```diff
@@ -0,0 +1,94 @@
...
+  X(-24, PLACE_SWAP_BLOB_TO_BUF_FAILED, R"(Place blob from swap space into \
+                                      buffering system failed.)") \
+  X(-25, HERMES_ERROR_MAX,           "Maximum supported HERMES errors.") \
+
+#define RETURN_ENUM(ID, NAME, TEXT) NAME = ID,
+#define RETURN_TEXT(ID, NAME, TEXT) case ID: return TEXT;
+
+typedef enum {
+    RETURN_CODES(RETURN_ENUM)
+} FuncStatus;
```
ChristopherHogan, line 48: I would suggest naming this enum `StatusCode`. Also, no need for a typedef, just `enum StatusCode {}`.

src/hermes_status.h
```diff
@@ -0,0 +1,94 @@
...
+ public:
+  /** Create an object representing success status. */
+  explicit Status(FuncStatus ret_code = HERMES_SUCCESS) : status_(ret_code) {}
+
+  /** Returns true if the call did exactly what the user expected */
+  bool Succeeded() const {
+    return (status_ == HERMES_SUCCESS);
+  }
+
+  /** Returns true if the call was success with ccateat. */
```
ChristopherHogan, line 60: Typo for 'caveat.'

benchmarks/dpe_bench.cc
```diff
@@ -165,7 +165,7 @@ int main(int argc, char **argv) {
   assert(placed_size == total_placed_size);
 
   // Aggregate placement schemas from the same target
-  if (!result) {
+  if (!result.Succeeded()) {
```
ChristopherHogan, line 168: I think this should be `result.Succeeded()` (without the `!`), right?

src/api/bucket.cc
```diff
@@ -30,8 +30,12 @@ Bucket::Bucket(const std::string &initial_name,
 
   if (IsBucketNameTooLong(name_)) {
     id_.as_int = 0;
+    throw std::length_error("Bucket name is too long (max length 256).");
```
ChristopherHogan, line 33: The max length message should use `kMaxBucketNameSize` instead of hardcoding `256`.

test/dpe_optimization_test.cc
```diff
@@ -34,7 +34,7 @@ void MinimizeIoTimePlaceBlob(std::vector<size_t> &blob_sizes,
                                           node_state.bytes_available,
                                           node_state.bandwidth, targets,
                                           schemas_tmp);
-  if (result) {
+  if (!result.Succeeded()) {
```
ChristopherHogan, line 37: Isn't `if (result.Failed())` more natural?

ChristopherHogan:
Can we remove the `try ... catch` blocks from the tests? An uncaught exception will print the `what()` automatically, so catching just to print `what()` is redundant.

src/hermes_types.h
jya-kmu: > Can we remove the `try ... catch` blocks from the tests? An uncaught exception will print the `what()` automatically, so catching just to print `what()` is redundant.

That's true. Most tests would be fine without `try ... catch`, but the mdm test will fail. Do you suggest we modify it? 
Function [TestMaxNameLength](https://github.com/HDFGroup/hermes/blob/kimmy/error_handling/test/mdm_test.cc#L168) is testing if bucket [`invalid_bucket`](https://github.com/HDFGroup/hermes/blob/kimmy/error_handling/test/mdm_test.cc#L173) can be created if the name is too long, if `throw` in the constructor is not caught in time, the test will fail. This will happen to [blob name test](https://github.com/HDFGroup/hermes/blob/kimmy/error_handling/test/mdm_test.cc#L185) as well.

src/hermes_status.h
jya-kmu: `StatusCode` sounds good.

benchmarks/dpe_bench.cc
jya-kmu: Right.

jya-kmu:
> Can we remove the `try ... catch` blocks from the tests? An uncaught exception will print the `what()` automatically, so catching just to print `what()` is redundant.

That's true. Most tests would be fine without `try ... catch`, but the mdm test will fail. Do you suggest we modify it? 
Function [TestMaxNameLength](https://github.com/HDFGroup/hermes/blob/kimmy/error_handling/test/mdm_test.cc#L168) is testing if bucket [`invalid_bucket`](https://github.com/HDFGroup/hermes/blob/kimmy/error_handling/test/mdm_test.cc#L173) can be created if the name is too long, if `throw` in the constructor is not caught in time, the test will fail. This will happen to [blob name test](https://github.com/HDFGroup/hermes/blob/kimmy/error_handling/test/mdm_test.cc#L185) as well.

src/hermes_types.h
jya-kmu: The reason is that `Status` is used by functions that belong to `src` as well, not just `src/api`. But it is true that the user will only interact with `hermes::api`.

ChristopherHogan:
In `TestMaxNameLength` I would suggest replacing 

```c++
try {
    std::string long_bucket_name(kMaxBucketNameSize + 1, 'x');
    hapi::Bucket invalid_bucket(long_bucket_name, hermes, ctx);
}
  catch (const std::runtime_error& e) {
    std::cout << "Standard exception: " << e.what() << std::endl;
}
  catch (const std::length_error& e) {
    std::cout << "Standard exception: " << e.what() << std::endl;
}
```
with
```c++
std::string long_bucket_name(kMaxBucketNameSize + 1, 'x');
bool threw_length_error = false;
try {
    hapi::Bucket invalid_bucket(long_bucket_name, hermes, ctx);
}
  catch (const std::length_error& e) {
    threw_length_error = true;
}
Assert(threw_length_error);
```

> This will happen to blob name test as well.

I don't see how the line you linked could throw an exception. If the `Blob` name is too long during a `Put`, we return a `StatusCode` instead of throwing an exception, right?

jya-kmu:
> In `TestMaxNameLength` I would suggest replacing
> 
> ```c++
> try {
>     std::string long_bucket_name(kMaxBucketNameSize + 1, 'x');
>     hapi::Bucket invalid_bucket(long_bucket_name, hermes, ctx);
> }
>   catch (const std::runtime_error& e) {
>     std::cout << "Standard exception: " << e.what() << std::endl;
> }
>   catch (const std::length_error& e) {
>     std::cout << "Standard exception: " << e.what() << std::endl;
> }
> ```
> 
> with
> 
> ```c++
> std::string long_bucket_name(kMaxBucketNameSize + 1, 'x');
> bool threw_length_error = false;
> try {
>     hapi::Bucket invalid_bucket(long_bucket_name, hermes, ctx);
> }
>   catch (const std::length_error& e) {
>     threw_length_error = true;
> }
> Assert(threw_length_error);
> ```
> 
> > This will happen to blob name test as well.
> 
> I don't see how the line you linked could throw an exception. If the `Blob` name is too long during a `Put`, we return a `StatusCode` instead of throwing an exception, right?

The replacement is good. Sorry, it is not `Blob` name is too long. I meant [this bucket](https://github.com/HDFGroup/hermes/blob/kimmy/error_handling/test/mdm_test.cc#L60). It can use a similar way you proposed.

src/hermes_types.h
jya-kmu: For the convenience of users, I will move it to `hermes::api`.

test/mdm_test.cc
```diff
@@ -221,7 +233,6 @@ int main(int argc, char **argv) {
   TestRenameBucket(hermes);
   TestBucketRefCounting(hermes);
   TestMaxNameLength(hermes);
-  TestGetRelativeNodeId();
```
ChristopherHogan, line 224: Why is this test deleted?

test/buffer_pool_test.cc
```diff
@@ -85,7 +85,7 @@ void TestGetBandwidths(hermes::SharedMemoryContext *context) {
   }
 }
 
-hapi::Status ForceBlobToSwap(Hermes *hermes, hermes::u64 id, hapi::Blob &blob,
+hermes::Status ForceBlobToSwap(Hermes *hermes, hermes::u64 id, hapi::Blob &blob,
```
ChristopherHogan, line 88: If you search the code for `hermes::Status` you'll find 7 hits. Could you please change those to `hapi::Status`? The reason it compiles is because `data_placement_engine.h` and `data_placement_engine.cc` both have `using api::Status;` statements in the `hermes` namespace.

src/hermes_status.h
```diff
@@ -0,0 +1,98 @@
...
+ public:
+  /** Create an object representing success status. */
+  explicit Status(StatusCode ret_code = HERMES_SUCCESS) : status_(ret_code) {}
+
+  /** Returns true if the call did exactly what the user expected */
+  bool Succeeded() const {
+    return (status_ == HERMES_SUCCESS);
+  }
+
+  /** Returns true if the call was success with cateat. */
```
ChristopherHogan, line 62: Still a typo. Should be "caveat."

test/mdm_test.cc
jya-kmu: Should be a mistake. I will get it back.

test/buffer_pool_test.cc
jya-kmu: Done!

src/hermes_status.h
ChristopherHogan: This can be merged after this typo is fixed.

ChristopherHogan (approved):

src/hermes_status.h
jya-kmu: Done

PR #149, opened 2021-03-10
Embed BucketID in internal representation of Blob names

### LINKED ISSUE #88: Add bucket id prefix to blob names
Created from JIRA [HERM-39](https://jira.hdfgroup.org/browse/HERM-39)

Different buckets could have blobs of the same name, but currently all blob names are kept in the same map. We need to distinguish blob names in different buckets by adding a bucket ID prefix to the name.

ChristopherHogan:
Fixes #88.

Prepends a `Blob`'s `BucketID` onto the internal representation of the `Blob` name. This allows each `Bucket` to have its own "namespace" in the MDM for `Blob` names. When the name is retrieved, the `BucketID` portion is discarded. I originally planned to stuff the raw bytes of the `BucketID` at the beginning of the name, but that causes null bytes to be interpreted as null terminators when the names are eventually treated as C strings in `stb_ds.h`. To get around this, I'm storing the hex representation of the `BucketID` as `char`s.  Luckily, since we generate all the numbers we have to convert from a hex string to a `u64`, we can get a 4.1x speedup over `std::stoull` by avoiding input sanitization and error checking.

* Renamed functions:
  * `GetBlobIdByName` -> `GetBlobId`
  * `GetBucketIdByName` -> `GetBucketId`
  * `GetVBucketdByName` -> `GetVBucketId`
* Adds `GetBucketIdFromBlobId()` needed for the adapter.

### DIFF
--- a/src/api/bucket.cc
+++ b/src/api/bucket.cc
@@ -169,7 +169,8 @@ Status Bucket::RenameBlob(const std::string &old_name,
     return ret;
   } else {
     LOG(INFO) << "Renaming Blob " << old_name << " to " << new_name << '\n';
-    hermes::RenameBlob(&hermes_->context_, &hermes_->rpc_, old_name, new_name, id_);
+    hermes::RenameBlob(&hermes_->context_, &hermes_->rpc_, old_name,
+                       new_name, id_);
   }
 
   return ret;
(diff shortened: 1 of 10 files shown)


ChristopherHogan:
The multi-node test is failing. I'm going to close this, rebase, test, and open a new PR.

PR #152, opened 2021-03-12
Don't remove swap blob metadata on failed BO Put.

ChristopherHogan:
* Swap blob entries were being incorrectly removed from the MDM in `PlaceBlob`.
* Disabled `hermes_stdio_low_buf_adapter_test` because it's failing intermittently. I'll troubleshoot it separately.

@hariharan-devarajan


jya-kmu (approved):
Does stdio test fail locally?

NP, found it on Github action.

PR #158, opened 2021-03-15
Hariharan/vbucket shared

### LINKED ISSUE #134: Sharing vBuckets across multiple processes.
We should be able to create a vBucket in global namespace. Then link blobs to it.
We should be able to create/initialize Traits in global namespace. 
And have function ptrs in shared memory. 
These function ptrs will be called on link/unlink of blob and attach/detach of trait on existing blobs.

hariharan-devarajan:
Fixed #134


test/vbucket_test.cc
```diff
@@ -0,0 +1,91 @@
...
+        for (int i = 0; i < num_blobs_per_rank; ++i) {
+          std::string blob_name =
+              "Blob_" + std::to_string(rank) + "_" + std::to_string(i);
+          auto exists = shared.Contain_blob(blob_name, bucket_name_temp);
+          assert(exists);
+        }
+      }
+    }
+    hermes->AppBarrier();
+    rank_bucket.Destroy(ctx);
```
ChristopherHogan, line 77: Can you add an `Assert(!shared.Contain_blob(...))` call here to make sure that the VBucket calls don't crash after their linked Blobs have been destroyed?

hariharan-devarajan: FIXED.

test/vbucket_test.cc
```diff
@@ -0,0 +1,91 @@
...
+    hapi::VBucket shared("shared_vbucket", hermes, false, ctx);
+    for (int i = 0; i < num_blobs_per_rank; ++i) {
+      std::string blob_name =
+          "Blob_" + std::to_string(app_rank) + "_" + std::to_string(i);
+      shared.Link(blob_name, bucket_name, ctx);
+    }
+    hermes->AppBarrier();
+    if (app_rank == 0) {
+      auto blob_ids = shared.GetLinks(ctx);
+      assert(blob_ids.size() == (unsigned long)app_size * num_blobs_per_rank);
```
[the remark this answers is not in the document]
hariharan-devarajan: FIXED.

ChristopherHogan:
Everything looks good here, but I still think it should be possible to have a `Blob` named "0". Is there a good reproducer for error?

hariharan-devarajan:
> Everything looks good here, but I still think it should be possible to have a `Blob` named "0". Is there a good reproducer for error?

For now no. Once I am done with other deadlines, I will create an issue for this and share it with you.

ChristopherHogan (approved):

PR #165, opened 2021-03-30
Call Close in Bucket destructor and Rename Close to Release

### LINKED ISSUE #132: The Bucket destructor should call Close

ChristopherHogan:
Closes #132.

* Rename `Bucket::Close` to `Bucket::Release` (requested by Anthony).
* Call `Bucket::Release` in `Bucket` destructor.

There are several patterns in the tests that require more precise control of `Bucket::Release`, so I've set it up so that the destructor only calls `Release` if the `Bucket` is still valid. Manual calls to `Release` are still possible and compatible with automatic calls in the destructor.

### DIFF
--- a/src/api/vbucket.cc
+++ b/src/api/vbucket.cc
@@ -115,7 +115,6 @@ Blob& VBucket::GetBlob(std::string blob_name, std::string bucket_name) {
   size_t blob_size = bkt.Get(blob_name, local_blob, ctx);
   local_blob.resize(blob_size);
   bkt.Get(blob_name, local_blob, ctx);
-  bkt.Close(ctx);
   return local_blob;
 }
 
--- a/test/bucket_test.cc
+++ b/test/bucket_test.cc
@@ -161,6 +161,43 @@ void TestPutOverwrite(std::shared_ptr<hapi::Hermes> hermes) {
   bucket.Destroy(ctx);
 }
 
+void TestCompressionTrait(std::shared_ptr<hapi::Hermes> hermes) {
+  hermes::api::Context ctx;
+  hermes::api::Status status;
+
+  const std::string bucket_name = "compression";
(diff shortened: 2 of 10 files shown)


test/bucket_test.cc
```diff
@@ -56,7 +56,6 @@ int compress_blob(hermes::api::TraitInput &input, hermes::api::Trait *trait) {
   size_t blob_size = bkt.Get(input.blob_name, blob, ctx);
   blob.resize(blob_size);
   bkt.Get(input.blob_name, blob, ctx);
-  bkt.Close(ctx);
```
jya-kmu, line 59: This bucket is not destroyed.

jya-kmu: There is the other bucket may that need to destroy in main #line 184

src/api/vbucket.cc
```diff
@@ -115,7 +115,6 @@ Blob& VBucket::GetBlob(std::string blob_name, std::string bucket_name) {
   size_t blob_size = bkt.Get(blob_name, local_blob, ctx);
   local_blob.resize(blob_size);
   bkt.Get(blob_name, local_blob, ctx);
-  bkt.Close(ctx);
```
jya-kmu, line 118: Do we replace `Close` with `Release` here?

test/bucket_test.cc
ChristopherHogan: I refactored the compression test into its own function so that it's easier to see which resources need to be freed. Thanks.

src/api/vbucket.cc
ChristopherHogan: No, because it's automatically called in the destructor.

jya-kmu (approved):

ChristopherHogan:
I'll wait to merge this until #158 is merged.

ChristopherHogan:
I'm going to merge this now because I don't know how long it will take to merge #158.

PR #167, opened 2021-04-02
Control PlacementPolicy at global, Bucket, and Put granularity

### LINKED ISSUE #162: Enable placement policies at the global, Bucket, and Put levels
Currently we can only change the placement policy (`Context::policy`) at the `Put` granularity. We need
1. A global default policy defined in the config file
2. A `Bucket` granularity policy.
3. A `Put` granularity policy.

ChristopherHogan:
Closes #162.

* Adds a `default_placement_policy` option to the configuration file. This becomes the default `policy` value for all instances of `api::Context`.
* Save the `Context` passed to a `Bucket`. The `Context::policy` in the `Bucket` overrides the global default.
* Add overloads of `Put` that don't take a `Context` argument. These versions simply use the `Context` stored in the `Bucket`.
* Passing a `Context` to a `Put` overrides the `Bucket`'s default `policy`.
* The full `Context` is preserved on swap blobs when the BO attempts to place them back into the hierarchy. Previously it used a default `Context`.

### DIFF
--- a/src/api/bucket.cc
+++ b/src/api/bucket.cc
@@ -25,7 +25,7 @@ namespace api {
 
 Bucket::Bucket(const std::string &initial_name,
                const std::shared_ptr<Hermes> &h, Context ctx)
-  : name_(initial_name), hermes_(h), placement_policy(ctx.policy) {
+  : name_(initial_name), hermes_(h), ctx_(ctx) {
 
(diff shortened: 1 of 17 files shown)


jya-kmu (approved):
Looks good.

PR #174, opened 2021-04-27
Hariharan/mpiio examples

hariharan-devarajan:
Contains all MPIIO examples using different APIs critical for MPIIO adapter.


adapter/test/mpiio/mpiio_adapter_test.cpp
```diff
@@ -0,0 +1,671 @@
...
+  info.existing_file_cmp =
+      fullpath.string() + "_ext_cmp" + "_" + std::to_string(getpid());
+  info.shared_new_file =
+      fullpath.string() + "_new_" + std::to_string(info.comm_size);
+  info.shared_existing_file =
+      fullpath.string() + "_ext_" + std::to_string(info.comm_size);
+  info.shared_new_file_cmp =
+      fullpath.string() + "_new_cmp_" + std::to_string(info.comm_size);
+  info.shared_existing_file_cmp =
+      fullpath.string() + "_ext_cmp_" + std::to_string(info.comm_size);
```
ChristopherHogan, line 104: Do these filenames need a pid appended?

ChristopherHogan: I see that they need a uniform name so that they can be shared by each process, but we still run into the issue where one person creates these files in `/tmp` on Ares and then the next person to run the tests gets permission errors.

hariharan-devarajan: Appended the USER variable to the filename.

ChristopherHogan (approved):

PR #176, opened 2021-05-06
Hariharan/mpiio adapter Syncronous Read/Write

hariharan-devarajan:
This PR implements the Sync APIs of MPI-IO adapter.


src/metadata_storage_stb_ds.cc
```diff
@@ -493,6 +493,49 @@ bool LocalDestroyBucket(SharedMemoryContext *context, RpcContext *rpc,
...
+  // this is slow
+  int ref_count = info->ref_count.load();
+  if (ref_count == 1) {
+    ChunkedIdList *blobs = &info->blobs;
+    BlobID *blobs_arr = (BlobID *)GetIdsPtr(mdm, *blobs);
+    for (u32 i = 0; i < blobs->length; ++i) {
+      blobs_arr[i] = blobs_arr[--blobs->length];
+      break;
+    }
+    ReleaseIdsPtr(mdm);
```
ChristopherHogan, line 513: I think instead of lines 507-513, we just need to call `FreeIdList(mdm, info->blobs)`, but only if `HasAllocatedBlobs(info)` is `true`.

adapter/src/hermes/adapter/mpiio/mpiio.cc
```diff
@@ -0,0 +1,1006 @@
...
+  size_t ret;
+  auto mdm = hermes::adapter::Singleton<MetadataManager>::GetInstance();
+  auto mapper = MapperFactory().Get(kMapperType);
+  int datatype_size;
+  MPI_Type_size(datatype, &datatype_size);
+  size_t total_size = datatype_size * count;
+  auto mapping = mapper->map(FileStruct(fp, existing.first.ptr, total_size));
+  size_t data_offset = 0;
+  auto filename = existing.first.st_bkid->GetName();
+  LOG(INFO) << "Mapping for read has " << mapping.size() << " mapping."
```
ChristopherHogan, line 165: Should this be "Mapping for write?"

adapter/src/hermes/adapter/mpiio/mpiio.cc
```diff
@@ -0,0 +1,1006 @@
...
+  for (const auto &item : mapping) {
+    hapi::Context ctx;
+    auto index = std::stol(item.second.blob_name_) - 1;
+    unsigned char *put_data_ptr = (unsigned char *)ptr + data_offset;
+    size_t put_data_ptr_size = item.first.size_;
+
+    if (item.second.size_ == kPageSize) {
+      LOG(INFO) << "Create or Overwrite blob " << item.second.blob_name_
+                << " of size:" << item.second.size_ << "." << std::endl;
+      if (item.second.size_ == kPageSize) {
```
ChristopherHogan, line 176: Since the `if` conditions on 173 and 176 are the same, doesn't that make lines 185-214 unreachable? In other words, if the condition on 173 is true, then 176 is also true, and we will never hit the `else` branches on 185 and 194.

adapter/src/hermes/adapter/mpiio/mpiio.cc
```diff
@@ -0,0 +1,1006 @@
...
+    data_offset += item.first.size_;
+  }
+  existing.first.ptr += data_offset;
+  existing.first.size = existing.first.size >= existing.first.ptr
+                            ? existing.first.size
+                            : existing.first.ptr;
+  mdm->Update(fp, existing.first);
+  ret = data_offset;
+  mpi_status->count_hi_and_cancelled = 0;
+  mpi_status->count_lo = ret;
```
ChristopherHogan, line 324: These fields only exist in MPICH, and are implementation specific. Here's a [relevant comment](https://github.com/open-mpi/ompi/blob/master/ompi/include/mpi.h.in#L431) from `MPI_Status` in OpenMPI. Also see section 3.2.5 of the [MPI standard](https://www.mpi-forum.org/docs/mpi-4.0/mpi40-rc-jun-21.pdf#page=79&zoom=100,118,46). Maybe there are API functions that let you set this information?

adapter/src/hermes/adapter/mpiio/mpiio.cc
```diff
@@ -0,0 +1,1006 @@
...
+/**
+ * Other functions
+ */
+int HERMES_DECL(MPI_File_sync)(MPI_File fh) {
+  int ret = -1;
+  if (IsTracked(&fh)) {
+    auto mdm = hermes::adapter::Singleton<MetadataManager>::GetInstance();
+    auto existing = mdm->Find(&fh);
+    if (existing.second) {
+      LOG(INFO) << "Intercept fflush." << std::endl;
```
ChristopherHogan, line 975: fflush -> MPI_File_sync

src/metadata_storage_stb_ds.cc
ChristopherHogan: I fixed this in #201.

adapter/src/hermes/adapter/mpiio/mpiio.cc
ChristopherHogan: Fixed in #201.

adapter/src/hermes/adapter/mpiio/mpiio.cc
ChristopherHogan: Commented the unreachable code out and created issue #203.

adapter/src/hermes/adapter/mpiio/mpiio.cc
ChristopherHogan: Created issue #202.

adapter/src/hermes/adapter/mpiio/mpiio.cc
ChristopherHogan: Fixed in #201.

PR #201, opened 2021-06-16
Replaces #176

ChristopherHogan:


ChristopherHogan:
This is Hari's PR #176, with a couple obvious fixes. I'll create issues for the remaining questions I have. For now the MPI-IO adapter only supports MPICH, but if you want to build Hermes with OpenMPI, you can just set `HERMES_ENABLE_ADAPTERS=OFF`.

jya-kmu (approved):

PR #207, opened 2021-06-22
Bug fixes and improvements related to Hermes VPIC proxy

ChristopherHogan:
* Adapter client finalization can optionally terminate daemon.
* `RoundRobin` DPE algorithm, `LocalGetTargets`, and `GetBandwidths` only considers targets with a non-zero number of buffers.
* Only initialize shared files (e.g., burst buffer files) on one rank.
* Support configurations with no RAM buffers.
* Add `is_shared_device` array to configuration.
* `RoundRobin` placement shouldn't fail if the `Target` list passed to it doesn't contain any instances of the "current" device.
* Add `Hermes::RunDaemon` for Hermes daemon programs.
* Replace `gethostbyname` with threadsafe version.
* MPI-IO adapter initialization and finalization only needs to happen in `MPI_Init` and `MPI_Finalize`, not when opening and closing files.


jya-kmu (approved):
Looks good!

Issue #213, opened 2021-07-08
Convert Stats arrays to maps


ChristopherHogan:
Each `Blob` statistic update involves a linear lookup. Once #212 is finished, we can convert the `BucketInfo::stats` array to a map.

ChristopherHogan:
Fixed in #219.

Issue #212, opened 2021-07-08
Refactor metadata storage


ChristopherHogan:
Goals:
1. Don't force user to specify metadata size.
2. Allow metadata to grow dynamically instead of choosing a constant size up front.

Tasks:
1. Split metadata into its own shared memory segment.
2. Remove all metadata parameters from Hermes config file.
3. Implement `Heap::realloc`.

Dynamically resizing shared memory segments is difficult because we don't have any natural synchronization points in a running Hermes app outside of initialization and finalization. Synchronization is required because each process must `remap` it's shared memory segment after one process has grown the file via `ftruncate`. We can avoid this growing and remapping by giving the segment an initial `max_metadata_size` (configurable). Then we just map the whole segment. Based on some testing I did, only the blocks of the segment that we actually read/write are brought into physical memory (page cache actually), and once we're out of physical memory, those blocks are reclaimed (and dirty pages are written to disk). Therefore, mapping a 2GiB file doesn't put any pressure on physical memory, and it allows us to realloc within the segment, giving us dynamic maps, without app wide synchronization.

PR #214, opened 2021-07-08
Basic Blob statistics and scoring

ChristopherHogan:
This is the initial infrastructure for `Blob` statistics and scoring. The details can be refined as we discuss more as a group.

* Tracks recency and frequency of accesses for each `Blob`.
* Adds `GetBlobScore` function.

**TODO**: #213. Stat lookups are currently slow (`O(n)`) because they are stored in arrays. Once #212 is finished, I can convert the stats into a `BlobID -> Stats` map.

### DIFF
--- a/src/metadata_storage_stb_ds.cc
+++ b/src/metadata_storage_stb_ds.cc
@@ -292,7 +292,7 @@ u32 AppendToChunkedIdList(MetadataManager *mdm, ChunkedIdList *id_list,
 u64 GetChunkedIdListElement(MetadataManager *mdm, ChunkedIdList *id_list,
                             u32 index) {
   u64 result = 0;
-  if (index < id_list->length) {
+  if (id_list->length && index < id_list->length) {
     u64 *head = GetIdsPtr(mdm, *id_list);
     result = head[index];
     ReleaseIdsPtr(mdm);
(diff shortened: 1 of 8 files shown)


src/metadata_storage_stb_ds.cc
```diff
@@ -272,23 +272,111 @@ void AllocateOrGrowIdList(MetadataManager *mdm, ChunkedIdList *id_list) {
...
+  }
 }
 
 void LocalAddBlobIdToBucket(MetadataManager *mdm, BucketID bucket_id,
                             BlobID blob_id) {
   BeginTicketMutex(&mdm->bucket_mutex);
   BucketInfo *info = LocalGetBucketInfoById(mdm, bucket_id);
-  AppendToChunkedIdList(mdm, &info->blobs, blob_id.as_int);
+  u32 index = AppendToChunkedIdList(mdm, &info->blobs, blob_id.as_int);
+  AppendToChunkedIdList(mdm, &info->stats, 0);
```
jya-kmu, line 378: What is the reason to call `AppendToChunkedIdList` twice with different last parameter?

jya-kmu (approved):
Please merge it if the question is not an issue

src/metadata_storage_stb_ds.cc
ChristopherHogan: The first call is to add the `BlobID` to the `Bucket`'s list of blobs. The second call is to add the blob statistics, which I treat as an ID here because it can be read as a `u64`. `info->stats[i]` represents the stats for `info->blobs[i]`.This will go away once I implement #213.

PR #282, opened 2021-11-05
Automatically set Config::num_targets == Config::num_devices

### LINKED ISSUE #259: `num_targets` should default to the value of `num_devices` in the config parser
For the beta release we want to just hide `num_targets`, since there is currently no clear distinction between it and `num_devices`. That means the config parser needs to set `num_targets` to the value of `num_devices` automatically.

ChristopherHogan:
Fixes #259.


jya-kmu (approved):

ChristopherHogan:
Test failures are expected and unrelated.

PR #301, opened 2021-11-19
[HERM] Expand critical sections in Heap functions

### LINKED ISSUE #286: Fix failing `Testhermes_stdio_adapter_mpi_2_` test
It fails periodically on the `master` branch. I haven't been able to reproduce it locally.

ChristopherHogan:
Hopefully this fixes #286. I could never reproduce the failing test locally, but I ran it on the CI machine over 20 times without it failing, which is encouraging because it used to fail about 70% of the time.


jya-kmu:
Have you tried `fsanitize=thread` to check race condition?

jya-kmu (approved):

ChristopherHogan:
> Have you tried fsanitize=thread to check race condition?

Unfortunately the thread sanitizer is only able to detect pthreads constructs. Since we have to allocate locks in shared memory we can't use the standard primitives, so the sanitizer is unaware of our locks.

PR #316, opened 2022-01-08
pubsub system

JaimeCernuda:
Shared files changed include:
adapter/CMakeLists.txt
adapter/test/CMakeLists.txt

Both to add the pubsub folder. 

I tried to change the CMakes and include following the format followed by other adapters. 
I included the BSD license into every file

I have compiled and run every one of the tests included.

Things left to do:
Explore performance in the fulls scale of Ares
Test for multi-threaded. I have only tested multiprocess tests


ChristopherHogan:
Can you fix the `cpplint` errors please? Click on the failed Github Action to see the details.

JaimeCernuda:
Updated PR with all checks passed.

adapter/pubsub/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
+project(StdioAdapter VERSION 0.1.0)
```
ChristopherHogan, line 1: Wrong project name.

adapter/pubsub/constants.h
```diff
@@ -0,0 +1,61 @@
+/* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * *
```
ChristopherHogan, line 1: I don't think you need this file. It is already present in `adapter/constants.h`.

adapter/pubsub/datastructures.h
```diff
@@ -0,0 +1,61 @@
...
+ * All rights reserved.                                                      *
+ *                                                                           *
+ * This file is part of Hermes. The full Hermes copyright notice, including  *
+ * terms governing use, modification, and redistribution, is contained in    *
+ * the COPYING file, which can be found at the top directory. If you do not  *
+ * have access to the file, you may request a copy from help@hdfgroup.org.   *
+ * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
+
+#ifndef HERMES_STDIO_ADAPTER_DATASTRUCTURES_H
+#define HERMES_STDIO_ADAPTER_DATASTRUCTURES_H
```
ChristopherHogan, line 14: Should be `HERMES_PUBSUB...`

adapter/pubsub/datastructures.h
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Standard header
+ */
+#include <ftw.h>
```
ChristopherHogan, line 19: What do you need this header for? I think it can be removed.

adapter/pubsub/metadata_manager.h
```diff
@@ -0,0 +1,134 @@
...
+ * have access to the file, you may request a copy from help@hdfgroup.org.   *
+ * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
+
+#ifndef HERMES_ADAPTER_METADATA_MANAGER_H
+#define HERMES_ADAPTER_METADATA_MANAGER_H
+
+/**
+ * Standard headers
+ */
+#include <ftw.h>
```
ChristopherHogan, line 19: Is this used?

adapter/pubsub/metadata_manager.h
```diff
@@ -0,0 +1,134 @@
...
+ * All rights reserved.                                                      *
+ *                                                                           *
+ * This file is part of Hermes. The full Hermes copyright notice, including  *
+ * terms governing use, modification, and redistribution, is contained in    *
+ * the COPYING file, which can be found at the top directory. If you do not  *
+ * have access to the file, you may request a copy from help@hdfgroup.org.   *
+ * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
+
+#ifndef HERMES_ADAPTER_METADATA_MANAGER_H
+#define HERMES_ADAPTER_METADATA_MANAGER_H
```
ChristopherHogan, line 14: Please add `PUBSUB` in the include guard.

adapter/pubsub/metadata_manager.h
```diff
@@ -0,0 +1,134 @@
...
+/**
+ * Internal headers
+ */
+#include "datastructures.h"
+
+using hermes::adapter::pubsub::ClientMetadata;
+
+namespace hermes::adapter::pubsub {
+/**
+ * Metadata manager for STDIO adapter
```
ChristopherHogan, line 33: STDIO -> PUBSUB

adapter/test/pubsub/pubsub_connection_test.cc
```diff
@@ -0,0 +1,29 @@
...
+  hermes::pubsub::mpiInit(argc, argv);
+
+  char *config_file = 0;
+  if (argc == 2) {
+    config_file = argv[1];
+  } else {
+    config_file = getenv(kHermesConf);
+  }
+
+  auto connect_ret = hermes::pubsub::connect(config_file);
```
ChristopherHogan, line 25: This test is redundant because all your other tests call `connect` and `disconnect`. Please remove it.

ChristopherHogan:
Can you replace the `assert`s in the tests with the version [here](https://github.com/HDFGroup/hermes/blob/chogan/bo_organize/test/test_utils.h#L59)? Otherwise they are compiled out in release builds.

adapter/test/pubsub/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+endforeach()
+
+foreach(program ${EXTENDED_PUBSUB_TESTS})
+    target_include_directories(${program} PUBLIC ${HERMES_ADAPTER_DIR})
+    target_include_directories(${program} PUBLIC ${HERMES_ADAPTER_TEST_DIR})
+endforeach()
+
+if(HERMES_INSTALL_TESTS)
+    set(PUBSUB_TESTS pubsub_connection_test pubsub_metadata_test pubsub_negative_connection_test pubsub_topic_test
+            pubsub_end_to_end_test pubsub_end_to_end_test_sync)
```
ChristopherHogan, line 30: Can you use `EXTENDED_PUBSUB_TESTS` here? Also, I don't see a test called `pubsub_negative_connection_test`.

adapter/pubsub/pubsub.cc
```diff
@@ -0,0 +1,192 @@
...
+    } else {
+      return hapi::Status(hermes::INVALID_FILE);
+    }
+  } catch (const std::exception& e) {
+    LOG(FATAL) << "Could not connect to hermes daemon" <<std::endl;
+    return hapi::Status(hermes::INVALID_FILE);
+  }
+}
+
+hapi::Status hermes::pubsub::connect() {
```
ChristopherHogan, line 45: Rather than duplicating all the code from above, can this function just call the other `connect`?

```C++
hapi::Status hermes::pubsub::connect() {
  char* hermes_config = getenv(kHermesConf);
  return hermes::pubsub::connect(hermes_config);
}
```

adapter/pubsub/pubsub.cc
```diff
@@ -0,0 +1,192 @@
...
+hapi::Status hermes::pubsub::attach(const std::string& topic) {
+  LOG(INFO) << "Attaching to topic: " << topic << std::endl;
+  hapi::Context ctx;
+  auto mdm = hermes::adapter::Singleton
+      <hermes::adapter::pubsub::MetadataManager>::GetInstance();
+  auto existing = mdm->Find(topic);
+  if (!existing.second) {
+    LOG(INFO) << "Topic not existing" << std::endl;
+    ClientMetadata stat;
+    // stat.ref_count = 1;
```
ChristopherHogan, line 85: If this line isn't needed, please delete it.

ChristopherHogan:
All of the tests need an `MPI_Finalize()` at the end.

adapter/test/pubsub/pubsub_end_to_end_test_sync.cc
```diff
@@ -0,0 +1,61 @@
...
+    for (const auto& data : full_data) {
+      auto publish_ret = hermes::pubsub::publish("test", data);
+      assert(publish_ret.Succeeded());
+    }
+
+    unsigned long num_messages = full_data.size() * comm_size;
+    std::pair<hapi::Blob, hapi::Status> subscribe_ret;
+    for (unsigned long i = 0; i < num_messages; i++) {
+      subscribe_ret = hermes::pubsub::subscribe("test");
+      assert(subscribe_ret.second.Succeeded());
```
ChristopherHogan, line 52: Shouldn't you also `assert` that the messages were read in the correct order?

adapter/test/pubsub/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+        pubsub_end_to_end_test pubsub_end_to_end_test_sync)
+set(EXTENDED_PUBSUB_TESTS ${PUBSUB_TESTS} pubsub_metadata_test)
+foreach(program ${PUBSUB_TESTS})
+    add_executable(${program} ${CMAKE_CURRENT_SOURCE_DIR}/${program}.cc)
+    add_dependencies(${program} hermes_pubsub)
+    target_link_libraries(${program} ${LIBRT} hermes MPI::MPI_CXX
+            $<$<BOOL:${HERMES_RPC_THALLIUM}>:thallium> hermes_pubsub)
+    target_compile_definitions(${program}
+            PRIVATE $<$<BOOL:${HERMES_RPC_THALLIUM}>:HERMES_RPC_THALLIUM>)
+    mpi_daemon(${program} 1 "" "" 1)
```
ChristopherHogan, line 20: It would be nice to have some tests that run more than 1 rank.

JaimeCernuda:
I think I have addressed all the requests. 

There are two things of note:
- On "Can you replace the asserts in the tests with the version here? Otherwise they are compiled out in release builds.". I found linking to the tests_utils.h on the hermes/test folder to be very ugly, instead, I made a copy of the file in hermes/adapeters/test/pubsub and trim it down to only contain the asset macro. If you find this solution not ideal, let me know.
- The second to last comment "Shouldn't you also assert that the messages were read in the correct order?". Originally this test was intended to be a distributed example, it would seem at some point I forgot to separate it into its own section on the CMake. That has been fixed, and now the test runs with two clients. As it has multiple client i disregarded the order of the messages looking more for the fact that 2 clients publish 3 messages each and that every client was capable of reading 6 total messages.

adapter/pubsub/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+set(HERMES_PUBSUB_ADAPTER_DIR ${HERMES_ADAPTER_DIR}/pubsub)
+
+set(PUBSUB_ADAPTER_PUBLIC_HEADER pubsub.h)
+set(PUBSUB_ADAPTER_PRIVATE_HEADER ${CMAKE_CURRENT_SOURCE_DIR}/metadata_manager.h
+        ${CMAKE_CURRENT_SOURCE_DIR}/datastructures.h
+        ${HERMES_ADAPTER_DIR}/constants.h
+        ${HERMES_ADAPTER_DIR}/singleton.h)
+set(PUBSUB_ADAPTER_SRC pubsub.cc metadata_manager.cc)
+
+# Add library hermes_stdio
```
ChristopherHogan, line 12: `hermes_stdio` -> `hermes_pubsub`. Or just remove the comment since it doesn't provide any information that's not already present in the line it comments on.

ChristopherHogan (approved):

Issue #50, opened 2020-12-07
Build on specific OS version(s) rather than using ubuntu-latest in GH actions


ChristopherHogan:
We want to avoid our CI suddenly building on a different OS version. Better to be explicit.

PR #367, opened 2022-03-03
Move docker files to folder and add CI step to publish docker images

### LINKED ISSUE #290: Make the Docker image creation a GitHub action
- Use the system MPI and point Spack at that (to prevent bloating the build time)

### LINKED ISSUE #291: Move the Dockerfiles to a subfolder
Less clutter at the top level! (?)

ChristopherHogan:
Closes #290. Closes #291.

This change enables automatically building docker images and pushing them to dockerhub every time we tag a new release. I've also added a way to force the images to be built and uploaded via the `build_docker_images` workflow dispatch option.


.github/workflows/main.yml
```diff
@@ -92,3 +94,38 @@ jobs:
         uses: coverallsapp/github-action@master
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
+  build-docker-images:
+    if: ${{ github.ref_type == 'tag' || github.event.inputs.build_docker_images }}
+    runs-on: ubuntu-20.04
```
hyoklee, line 99: Use `ubuntu-latest` so we don't have to update it when a new Ubuntu release comes out.

hyoklee (commented):
How about having a separate Action for Docker image publishing since you want to run it only when there's a new tag?
main.yml seems too crowded.

.github/workflows/main.yml
ChristopherHogan: I used to have it set to `ubuntu-latest` but when it updated, it broke our builds. I'd rather have explicit control over updating so that we can do it when its convenient to us rather than having to put out fires. See #50 and #51.

ChristopherHogan:
> How about having a separate Action for Docker image publishing since you want to run it only when there's a new tag?
main.yml seems too crowded.

Done.

Issue #323, opened 2022-01-18
Can't properly initialize and shutdown GLOG


ChristopherHogan:
#273 added proper initialization and finalization of the GLOG library to silence "GLOG not initialized" warnings. However, explicitly shutting down GLOG causes problems with the Hermes VFD.

## The Problem

In `Hermes::Finalize` we call `ShutdownGoogleLogging`. In the VFD, `Hermes::Finalize` is called from `H5FD__hermes_term`, which is a function registered with `atexit(3)`. It appears that `atexit(3)`-registered functions run [after static global objects are destroyed](https://community.ibm.com/community/user/ibmz-and-linuxone/blogs/fang-lu2/2020/03/24/be-careful-when-using-the-atexit-routine-with-static-objects). So, GLOG's static objects are destroyed, then we get to `Hermes::Finalize` and `ShutdownGoogleLogging`, where it tries to access those static objects and segfaults. 

## The Solution

Use `IsGoogleLoggingInitialized` to see whether or not GLOG needs to be shut down.

## The Problem with That

`IsGoogleLoggingInitialized` is only available in GLOG `v.0.5.0`.

## The Solution

Require GLOG `v0.5.0`.

## The Problem with That

OR-Tools is [hardcoded](https://github.com/google/or-tools/blob/v7.7/cmake/dependencies/CMakeLists.txt#L89) to required GLOG `v0.4.0`. In newer (> 8.1) versions of OR-Tools they re-implemented their own version of GLOG, so we could update OR-Tools and use GLOG `v0.5.0` in Hermes. However, something about the OR-Tools GLOG re-implementation [makes it impossible](https://github.com/google/or-tools/issues/2378) to use in a library that depends on GLOG (i.e., Hermes). Apparently they're waiting for Abseil developers to update to update their flags API (since GFLAGS is now deprecated?).

## The Solution

**Don't** properly init GLOG. **Don't** properly shutdown GLOG. Live with the warnings. Once [this issue](https://github.com/google/or-tools/issues/2378) is fixed we can update to a newer OR-Tools and GLOG `v0.5.0`. Alternatively we can replace OR-Tools and avoid all these headaches. See #321.

PR #370, opened 2022-03-09
Prefer ORtools 7.7 by default

hyoklee:


ChristopherHogan:
From my understanding, our `ortools` spack package isn't a proper spack package in the sense that its dependencies are not satisfied by other spack packages, but rather are built from scratch as part of the `make third_party` command [here](https://github.com/HDFGroup/hermes/blob/master/ci/hermes/packages/gortools/package.py#L47). This command runs [this makefile](https://github.com/google/or-tools/blob/v7.7/makefiles/Makefile.third_party.unix.mk), which checks out sources of all the dependencies and builds them from scratch. `Coinutils` is included as part of our `ortools` spack package, but spack isn't aware of it. If you add `coinutils` as a dependency, spack will build and install `coinutils`, but so will `make third_party`, and the spack-installed version will likely be ignored. What prompted you to open this PR?

ChristopherHogan:
From my understanding, our `ortools` spack package isn't a proper spack package in the sense that its dependencies are not satisfied by other spack packages, but rather are built from scratch as part of the `make third_party` command [here](https://github.com/HDFGroup/hermes/blob/master/ci/hermes/packages/gortools/package.py#L47). This command runs [this makefile](https://github.com/google/or-tools/blob/v7.7/makefiles/Makefile.third_party.unix.mk), which checks out sources of all the dependencies and builds them from scratch. `Coinutils` is included as part of our `ortools` spack package, but spack isn't aware of it. If you add `coinutils` as a dependency, spack will build and install `coinutils`, but so will `make third_party`, and the spack-installed version will likely be ignored. What prompted you to open this PR?

hyoklee:
@ChristopherHogan , I tried to port Hermes on macOS Monterey via Spack and gortools failed due to coinutils dependency:

```
 configure: error: Required package CoinUtils not available.
     5781    -> compiling obj/static/O.darwin.x86_64.gnu.opt/lib/scip/scip_probing.o
```

ChristopherHogan:
What command produced this error? I'm confused because our ortools package doesn't seem to have a `configure` step.

hyoklee:
It occurs when you run without specific version like 7.7: `spack install gortools`

ChristopherHogan:
I see. We have to use version `7.7` of `ortools` because of #323.

hyoklee:
How about removing `:` from `7.7:`  in `gortools/package.py`?

ChristopherHogan:
Doesn't efaa4283448a8b3  just mean "only depend on `cmake` if we're installing `ortools@7.7`? If you're suggesting that the package should only allow building version `7.7`, then I think we need to remove all the `version()` calls except for `7.7`.

hyoklee:
Spack automatically picks up whatever highest version available when ':' is used.
Yes, removing all other versions will be good to avoid confusion.

ChristopherHogan (approved):

PR #402, opened 2022-04-09
Remove empty line in enum.

hyoklee:


ChristopherHogan:
I actually like to leave this extra space in on purpose for every enum because it serves as a reminder that the `<enumName>_Count` value must always come last.

hyoklee:
OK.

PR #406, opened 2022-04-14
Fixes for using Thallium +cereal

ChristopherHogan:
Cereal requires all objects passed to `save` functions to be `const`. I also had to move the serialization functions for `BoPriority` into the `#ifndef THALLIUM_USE_CEREAL` block because the default thallium serialization doesn't handle enums automatically.

### DIFF
--- a/src/rpc_thallium.h
+++ b/src/rpc_thallium.h
@@ -157,10 +157,9 @@ void load(A &ar, MapType &map_type) {
   ar.read(&val, 1);
   map_type = (MapType)val;
 }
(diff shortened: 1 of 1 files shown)


ChristopherHogan:
Yes, I planned on doing that in a separate PR.

ChristopherHogan:
The reason I'm updating CI in a separate PR is because PNNL is waiting for this fix and I wanted to get it to them as quickly as possible. Updating CI could potentially take another 4-6 hours because changes in the `ci` folder trigger a rebuild of all dependencies, which extends the build time to 90 minutes.

PR #416, opened 2022-04-27
Add test suite for the Hermes VFD

### LINKED ISSUE #378: Find a way to correctly shut down Hermes in the VFD
In the Hermes VFD, Hermes is finalized in the `H5FD__hermes_term` function, which is registered by the hdf5 library with the `atexit` function. Therefore, Hermes is finalized when the `main` function returns. This creates a problem with MPI because `MPI_Finalize` is called before returning from main, but `H5FD__hermes_term` needs MPI to correctly shut everything down. The only solution I can think of is to implement `PMPI_Finalize` in the Hermes VFD and `LD_PRELOAD` it so we can do our shutdown before finalizing MPI.

### LINKED ISSUE #385: The VFD doesn't work with HDF5 apps in C
There are some incompatibilities with how HDF5, the VFD, and GLOG interact. The HDF5 library terminates the Hermes VFD in a function registered with `atexit`. So the flow is like this:

```C
int main() {
  // My HDF5 app
  return 0; 
  // Hermes is shutdown here
}
```
The problem is that static variables in GLOG are destroyed after `return 0` but before Hermes shuts down. Then Hermes tries to use GLOG during shutdown, and it crashes. Static variable destruction and `atexit` functions are run in the reverse order in which they were initialized/called. Currently the flow looks like this:

1. The HDF5 library registers `H5FD__hermes_term` with `atexit`
2. Hermes is initialized, which initializes GLOG

Destruction happens in reverse order, so GLOG static objects are destroyed, then `H5FD__hermes_term` runs. The strange thing is that this only happens when the application is compiled as a C program. If I compile as C++, it works, although I think static variable initialization/destruction order can be randomized, so it may not always work.

To get the correct behavior we need to initialize Hermes before HDF5 registers `H5FD__hermes_term`. I tested this theory by calling `dlopen("libhermes.so")` at the top of the `main` function in `ior`. It worked, which confirms the theory, but I don't see a way to make that happen for arbitrary apps. We need a VFD hook that runs before `atexit` is called, which happens when HDF5 initializes.

### LINKED ISSUE #386: Create a test suite for the VFD
Use the posix and stdio adapter tests as a guide. Ideally we can reduce some code duplication in the tests and factor out a test module that all adapters can use.

ChristopherHogan:
Closes #378. Closes #385. Closes #386.
* I tried to follow the same structure as the other adapter tests so that we can eventually reduce code duplication in the adapters.
* Fixed several bugs in the VFD.
* Since we need to intercept HDF5 and MPI initialization and finalization, `libhdf5_hermes_vfd.so` must be `LD_PRELOAD`ed in addition to setting `HDF5_DRIVER=hermes`.


adapter/CMakeLists.txt
```diff
@@ -21,6 +21,13 @@ endif()
 
 if(HERMES_ENABLE_VFD)
   if(HERMES_ENABLE_WRAPPER)
+    # The Hermes VFD cannot be installed into the same directory as libhermes.so
+    # because of the way dynamic plugin loading works in HDF5. When searching
+    # for a VFD, HDF5 will dlopen and dlclose every shared library in
+    # HDF5_PLUGIN_PATH looking for a valid VFD, and when it goes through this
+    # process with libhermes.so it resets some statically initialized variables.
```
hyoklee, line 28: I wish you could split the setence for clarity.
```
When searching for a VFD, HDF5 will dlopen and dlclose every shared library in
HDF5_PLUGIN_PATH looking for a valid VFD. When HDF5 goes through this
process with libhermes.so, it resets some statically initialized variables.
```

adapter/test/vfd/CMakeLists.txt
```diff
@@ -4,36 +4,64 @@ set(hermes_vfd_tests
...
+  set_property(TEST ${test_name} APPEND
+    PROPERTY ENVIRONMENT HDF5_DRIVER=hermes)
+  set_property(TEST ${test_name} APPEND
+    PROPERTY ENVIRONMENT LD_PRELOAD=${LIBASAN_PATH}:$<TARGET_FILE:hdf5_hermes_vfd>)
+  set_property(TEST ${test_name} APPEND
     PROPERTY ENVIRONMENT LSAN_OPTIONS=suppressions=${CMAKE_SOURCE_DIR}/test/data/asan.supp)
-endforeach()
+endfunction()
+
+add_test(NAME "TestVfd" COMMAND hermes_vfd_test --reporter compact -d yes)
```
hyoklee, line 50: Is there a reason for using Vfd instead of VFD?

adapter/test/vfd/CMakeLists.txt
```diff
@@ -4,36 +4,64 @@ set(hermes_vfd_tests
...
+set_vfd_test_properties("TestVfd")
+set_property(TEST "TestVfd" APPEND
+  PROPERTY ENVIRONMENT HDF5_DRIVER_CONFIG=true\ 65536)
```
hyoklee, line 53: What is `\ 65536` for after `true`?

adapter/test/vfd/CMakeLists.txt
```diff
@@ -4,36 +4,64 @@ set(hermes_vfd_tests
...
-    PROPERTY ENVIRONMENT HDF5_PLUGIN_PATH=${CMAKE_RUNTIME_OUTPUT_DIRECTORY})
-  set_property(TEST "TestVFDWrite" APPEND
-    PROPERTY ENVIRONMENT HDF5_DRIVER_CONFIG=true\ 1024)
-  set_property(TEST "TestVFDWrite" APPEND
-    PROPERTY ENVIRONMENT HDF5_DRIVER=hermes)
+  set(IOR_TEST_NAME "TestVfdIor")
+  add_test(NAME ${IOR_TEST_NAME} COMMAND ${IOR_EXE} -a HDF5 -w -W -r -R)
+  set_vfd_test_properties(${IOR_TEST_NAME})
+  set_property(TEST ${IOR_TEST_NAME} APPEND
+    PROPERTY ENVIRONMENT HDF5_DRIVER_CONFIG=true\ 262144)
```
hyoklee, line 66: What is `\ 262144` for?

adapter/test/vfd/hermes_vfd_basic_test.cc
```diff
@@ -0,0 +1,718 @@
...
+using hermes::adapter::vfd::test::MuteHdf5Errors;
+
+/** Returns a number in the range [1, upper_bound] */
+static inline size_t Random1ToUpperBound(size_t upper_bound) {
+  size_t result = ((size_t)GenNextRandom() % upper_bound) + 1;
+
+  return result;
+}
+
+/** Returns a string in the range ["0", "upper_bound") */
```
hyoklee, line 22: Isn't it `"upper_bound-1"`?

hyoklee (approved):

adapter/CMakeLists.txt
ChristopherHogan: Done.

adapter/test/vfd/CMakeLists.txt
ChristopherHogan: 1. It's easier to type.
2. I tend to use that convention because when you capitalize an acronym in camel case the acronym runs into the next word and is harder to read in my opinion. For example, in `VFDTest` the D and T run together whereas `VfdTest` has a clear separation between the two words.

adapter/test/vfd/CMakeLists.txt
ChristopherHogan: 65 KiB page size. If you're curious about the `\`, if I pass the config string as `"true 65536"` then CMake retains the quotes in the value of the environment variable. Using `\` escapes the space and sets the value of the variable to `true 65536` without the quotes.

adapter/test/vfd/CMakeLists.txt
ChristopherHogan: 256 KiB page size.

adapter/test/vfd/hermes_vfd_basic_test.cc
ChristopherHogan: I'm using interval notation. The `[` includes 0 on the left, but the `)` on the right means up to but not including `upper_bound`.

PR #424, opened 2022-05-26
Close #321.

### LINKED ISSUE #321: Replace OR-Tools
Given how many build issues we've encountered with OR-Tools, we should look at some alternative linear solver libraries. The only code changes involved would be to re-implement `hermes::MinimizeIoTimePlacement`.

hyoklee:
Replace OR-TOOLS with GLPK.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -270,6 +270,22 @@ if(HERMES_RPC_THALLIUM)
   endif()
 endif()
 
+# GLOG
+find_package(glog REQUIRED)
+if(glog_FOUND)
+  message(STATUS "found GLOG at ${glog_DIR}")
+endif()
+find_path(GLOG_INCLUDE_DIRS NAME glog PATH_SUFFIXES include/)
+include_directories(${GLOG_INCLUDE_DIRS})
+
+# GFLAGS
+find_package(gflags REQUIRED)
+if(gflags_FOUND)
+  message(STATUS "found GFLAGS at ${gflags_DIR}")
+endif()
+find_path(GFLAG_INCLUDE_DIRS NAME gflags PATH_SUFFIXES include/)
+include_directories(${GFLAG_INCLUDE_DIRS})
+
 # GOTCHA
 if(HERMES_INTERCEPT_IO)
   find_package(gotcha REQUIRED)
--- a/src/data_placement_engine.cc
+++ b/src/data_placement_engine.cc
@@ -13,13 +13,12 @@
 #include "data_placement_engine.h"
 
 #include <assert.h>
+#include <glpk.h>
 #include <math.h>
 
 #include <utility>
 #include <random>
 #include <map>
-
-#include "ortools/linear_solver/linear_solver.h"
 
 #include "hermes.h"
 #include "hermes_types.h"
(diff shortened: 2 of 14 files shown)


test/CMakeLists.txt
```diff
@@ -134,7 +134,7 @@ LSAN_OPTIONS=suppressions=${CMAKE_CURRENT_SOURCE_DIR}/data/asan.supp)
 #------------------------------------------------------------------------------
 add_executable(bo buffer_organizer_test.cc)
 target_link_libraries(bo ${LIBRT} hermes MPI::MPI_CXX
-  $<$<BOOL:${HERMES_RPC_THALLIUM}>:thallium>)
+  $<$<BOOL:${HERMES_RPC_THALLIUM}>:thallium> glog::glog gflags)
```
ChristopherHogan, line 137: Why does this test depend on gflags?

CMakeLists.txt
```diff
@@ -270,6 +270,22 @@ if(HERMES_RPC_THALLIUM)
...
+# GLOG
+find_package(glog REQUIRED)
+if(glog_FOUND)
+  message(STATUS "found GLOG at ${glog_DIR}")
+endif()
+find_path(GLOG_INCLUDE_DIRS NAME glog PATH_SUFFIXES include/)
+include_directories(${GLOG_INCLUDE_DIRS})
+
+# GFLAGS
+find_package(gflags REQUIRED)
```
ChristopherHogan, line 282: I would prefer to make gflags an optional dependency, or remove it altogether. I don't use it for anything, and glog works fine without it.

src/data_placement_engine.cc
```diff
@@ -272,103 +266,141 @@ Status MinimizeIoTimePlacement(const std::vector<size_t> &blob_sizes,
...
+    glp_set_row_bnds(lp, i+1, GLP_FX, 1.0, 1.0);
 
     // TODO(KIMMY): consider remote nodes?
     for (size_t j {0}; j < num_targets; ++j) {
+      int ij = i * num_targets + j + 1;
       std::string var_name {"blob_dst_" + std::to_string(i) + "_" +
                             std::to_string(j)};
-      blob_fraction[i][j] = solver.MakeNumVar(0.0, 1, var_name);
-      blob_constrt[num_constrts+i]->SetCoefficient(blob_fraction[i][j], 1);
+      // total umber of variables = number of blobs * number of targets.
```
ChristopherHogan, line 294: 'umber' -> 'number'

ChristopherHogan (changes requested):

CMakeLists.txt
hyoklee: I think glog depends on gflags.
```
/home/runner/work/hermes/hermes/spack/opt/spack/linux-ubuntu20.04-x86_64_v4/gcc-10.3.0/glog-0.4.0-jcohf3eh7cax7d6742d6owabbwdj6464/include/glog/logging.h:86:10: fatal error: gflags/gflags.h: No such file or directory
   86 | #include <gflags/gflags.h>
      |          ^~~~~~~~~~~~~~~~~
```

ChristopherHogan: You can disable the gflags dependency [via cmake](https://github.com/google/glog/blob/master/CMakeLists.txt#L41). Unfortunately it looks like the [glog spack package](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/glog/package.py) doesn't expose a variant without gflags. It hard-codes the dependency. I guess we're stuck with it for now.

ChristopherHogan (approved):

PR #437, opened 2022-07-26
Update to Catch2 version 3.0.1.

### LINKED ISSUE #434: Compilation fails when Catch2 3.0.1 is used.
Spack now uses version 3.0.1 for Catch2 by default.

hyoklee:
Spack has Catch2 package. It uses 3.0.1 version by default.


ChristopherHogan:
Could you use more descriptive titles for PRs please? When I auto-generate release notes it uses the PR title, and "Close 434" isn't very informative in release notes. I'd rather see something like "Update to catch2 3.0.1."

ChristopherHogan:
We should probably also add `3.0.0` as the minimum required version in the `find_package(Catch2 ...)` call of `CMakeLists.txt`.

ChristopherHogan:
I think they moved from a header only library to a normal library in version 3, so all the tests need to be linked with catch2 now.

ChristopherHogan (approved):

Issue #497, opened 2023-02-28
 error: the address of ‘hermes::Config::num_slabs’ will never be NULL [-Werror=address]


hyoklee:
gcc-**12.1.0** on Ubuntu latest  can't build config_parser.cc:

```
/home/runner/work/hermes/hermes/src/config_parser.cc:100:25: error: the address of ‘hermes::Config::num_slabs’ will never be NULL [-Werror=address]
  100 |   if (config->num_slabs == 0) {
      |       ~~~~~~~~~~~~~~~~~~^~~~
In file included from /home/runner/work/hermes/hermes/src/communication.h:16,
                 from /home/runner/work/hermes/hermes/src/buffer_pool.h:25,
                 from /home/runner/work/hermes/hermes/src/buffer_pool.cc:13:
/home/runner/work/hermes/hermes/src/hermes_types.h:264:7: note: ‘hermes::Config::num_slabs’ declared here
  264 |   int num_slabs[kMaxDevices];
      |       ^~~~~~~~~
```

PR #505, opened 2023-03-22
fix: gcc 12.1.0 compilation error (#497)

hyoklee:
* Check valid inputs more carefully.
* Log what configuration file is being used.


lukemartinlogan:
I've integrated this into the v1.0 release. Closing to avoid merge conflicts.

Issue #506, opened 2023-03-22
gcc 12 generates warnings for simple_io.cc


hyoklee:
```
/home/runner/work/hermes/hermes/adapter/test/posix/simple_io.cc: In function ‘int main(int, char**)’:
/home/runner/work/hermes/hermes/adapter/test/posix/simple_io.cc:93:12: warning: ignoring return value of ‘ssize_t write(int, const void*, size_t)’ declared with attribute ‘warn_unused_result’ [-Wunused-result]
   93 |       write(fd, buf, block_size);
      |       ~~~~~^~~~~~~~~~~~~~~~~~~~~
/home/runner/work/hermes/hermes/adapter/test/posix/simple_io.cc:96:11: warning: ignoring return value of ‘ssize_t read(int, void*, size_t)’ declared with attribute ‘warn_unused_result’ [-Wunused-result]
   96 |       read(fd, buf, block_size);
      |       ~~~~^~~~~~~~~~~~~~~~~~~~~
```

lukemartinlogan:
Include checks for read and write now.

PR #507, opened 2023-03-22
fix: gcc-12 compiler warnings (#506)

hyoklee:
This patch replaces __fxstat() to fstat() for Ubuntu 22.04.


lukemartinlogan:
I fixed this in the v1.0 PR

